PR #1, opened 2022-07-21
Update dependencies

larsnaesbye:
Updating the dependencies to newest.

I checked that the warnings are the same as before.


larsnaesbye:
Retracting this as I should have tested more thoroughly first.

PR #7, opened 2022-07-26
Add cargo audit check

larsnaesbye:
Check for known CVEs in crates.


larsnaesbye:
I like that it finds CVEs. Question is if a build should fail upon a CVE, or just inform,

larsnaesbye:
Maybe it’s better to run on Release.

gwbres:
@larsnaesbye 

Sorry I broke the pipeline for a bit.. 

I tested it on my side, but your script needs an extra `cargo install cargo-audit` to make the tool available, 
can you simply add this line prior invoking it?
a new submission will also fix the "checks" at the same time, due to my mistake

gwbres (approved):
Thank you for taking the time to declare the trigger event.
I'll add the "allow-failure" option to avoid this from blocking the pipeline though

PR #16, opened 2022-08-31
Show help if no parameters

### LINKED ISSUE #15: Calling `rinex-cli`without parameters panics
I tried not supplying any parameters to see if I got some help, but instead it panics:

```bash
lars@machine debug % ./rinex-cli        
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', rinex-cli/src/main.rs:432:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So I guess it should either abort with an error message or default to displaying a help - which the `clap` function `Command::arg_required_else_help` (2.x) or `arg_required_else_help`(3.x) can do :-)

larsnaesbye:
Letting `clap` display the help information when no parameter is given to `rinex-cli`

Fixes #15

### DIFF
--- a/rinex-cli/src/main.rs
+++ b/rinex-cli/src/main.rs
@@ -3,6 +3,7 @@
 //! Based on crate <https://github.com/gwbres/rinex>     
 //! Homepage: <https://github.com/gwbres/rinex-cli>
 use clap::App;
+use clap::AppSettings;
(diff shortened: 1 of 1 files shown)


gwbres (approved):
nice feature, thank you

PR #22, opened 2022-09-05
Add NAV and OBS testdata from same epoch

larsnaesbye:


larsnaesbye:
Apparently one file fails on gzip? I have no problem unpacking it locally in both Linux and Windows...

gwbres (approved):

gwbres:
> Apparently one file fails on gzip? I have no problem unpacking it locally in both Linux and Windows...

Apparently it's complaining about the `gzip` header content, for both new Observation files. Indeed `gzip -d` also worked on my side.

gwbres:
I'm not worried about this error, we'll figure it out.
I dropped a message on the decompression library this crate is currently using. I just demonstrated the two files are clearly gzip compressed and not anything else.

Thank you for the data, I will work on the double diff operation in the next few days

larsnaesbye:
Maybe this thread can help? 
https://users.rust-lang.org/t/flate2-decompress-is-returning-an-empty-slice/81931/8

My good colleague @busstoptaktik pointed me in the direction of it.

PR #34, opened 2022-09-26
Update Rust edition to 2021

larsnaesbye:
Since we're requiring Rust 1.61 anyway, we might as well use the 2021 edition introduced in 1.56.


gwbres (approved):

gwbres:
> Since we're requiring Rust 1.61 anyway, we might as well use the 2021 edition introduced in 1.56.

absolutely

PR #45, opened 2022-11-07
Epoch: upgrade to Hifitime

gwbres:
Upgrade `Epoch` to make use of [Hifitime::Epoch](https://github.com/nyx-space/hifitime/blob/master/src/epoch.rs)

:warning: files generated prior Jan 1 2000 with years encoded on two digits, like OBS V2 and NAV V2 get falsely shifted to 21st century. For example, "95" becomes 2095. Other than that, data is parsed correctly

- [x] Get rid of `chrono` dependency
- [x] Subsecond sample rates are now supported
- [x] Timestamp precision limit becomes 1 ns
  - [x] Observation timestamps (100ns precision)
  - [x] Navigation V2 (100ms precision)
- [x] Epoch formatting
  - [x] Obs V2
  - [x] Obs V3
  - [x] NAV V2
  - [x] NAV V3
  - [x] Others    
- [x] Take advantage of [Hifitime::TimeScale](https://github.com/nyx-space/hifitime/blob/master/src/timescale.rs)
  - [x] a RINEX timescale shifting method is introduced
  - [x] Clock RINEX description can take advantage of timescales too
- [x] [Leap second](https://github.com/nyx-space/hifitime/blob/master/src/timescale.rs#L124) support becomes natural
- [x] processing::Decimation 
  - [x] use new structures
  - [x] improve CI/CD, API, doc
- [x] `Rinex::epoch_intervals` 
  - [x] a new histogram analysis is introduced
  - [x] CI/CD 
  - [ ] propose in CLI application
- [x] (non related) NAV issues were spotted
  - [x] fixed a 1 byte offset when parsing epoch @ NAV 
  - [x] SBAS orbits badly identified: fixed
  - [x] `FNAV` frame identification missing: fixed
  - [x] `INAV` frame identification missing: fixed
- [x] (non related) CRINEX1/3 decompression issue appeared.
Omitted compressed observations were missed due to detection of such cases not smart enough
   - [x] fix it 
   - [x] add 2 thorough CRINX1 tests
   - [x] add 2 thorough CRINX3 tests
- [x] Update CLI/plot to `Hifitime::Epoch`
- [x] Pass CI/CD
- [x] Confirm PR


gwbres:
@larsnaesbye 

:warning: with this new feature, YYYY < 2000 is no longer supported.  

 `hifitime::from_gregorian_utc(YYYY, ...)`  needs a plausible YYYY value not to fail.

 When a two digit YY is detected 
 * when parsing compression timestamp
 * when parsing an epoch timestamp
 * basically parsing any timestamp

we evaluate YYYY = YY +2000 for the method to work.

I'll leave it like this, considering it's not very inconvenient to the end user. They will probably use all of this in a modern context.

The immediate consequence is we have fewer Meteo test resources for instance, I will upload a few more

gwbres:
Oddly enough, Navigation V3 timestamps precision is limited to 1sec, while V2 supported 0.1s.  
Crate supports all combinations correctly

PR #48, opened 2022-11-16
Python bindings

larsnaesbye:
- [ ] expose basic structures like `epoch::EpochFlag` or `observation::LliFlags` 
- [ ] modules
- [ ] convert record structures (`Vec<V>` or `Map<K, V>`) 
- [ ] we need a solution for the SBAS nested enum contained in Constellation, `pyclass` auto derivation does not allow that
- [ ] provide meaningful examples in python
- [ ] wheel auto generation and release
- [ ] readme/docs


rinex/Cargo.toml
```diff
@@ -35,7 +35,8 @@ geo = { version = "0.23.0", optional = true }
 wkt = { version = "0.10.0", default-features = false, optional = true }
 serde = { version = "1.0", optional = true, default-features = false, features = ["derive"] }
 flate2 = { version = "1.0.24", optional = true, default-features = false, features = ["zlib"] }
-hifitime = { version = "3.7.0", features = ["serde", "std"] }
+hifitime = { version = "3.7.0", features = ["serde", "std", "python"] }
```
gwbres, line 38: @larsnaesbye  
I am not advanced enough into Cargo, but the "python" features for "hifitime" should only be activated if our "pyo3" feature is activated

larsnaesbye: @gwbres
I am also unsure how to do this type of conditional feature activation.

rinex/src/epoch/flag.rs
```diff
@@ -10,10 +10,14 @@ pub enum Error {
...
 }
 
+#[cfg(feature = "pyo3")]
+use pyo3::prelude::*;
+
 /// `EpochFlag` validates an epoch, 
 /// or describes possible events that occurred
 #[derive(Copy, Clone, Debug)]
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "pyo3", pyclass)]
```
gwbres, line 20: @larsnaesbye ,
let's start easy with basic structures that do not have any dependents.

We'll have to solve several issues before we can derive `pyclass`  for `Rinex`.

We can't derive `pyclass`  for `Constellation` because Pyo3 does not allow nested enums like we have for SBAS description for example. Hopefully this is an open topic on their side? i don't know.

Then we need to describe the conversion for "complex" structures. Mainly: all record types and vectorized stuff. Apparently this is doable! but I just don't understand how to do it. An auto derivation method does not seem to exist. I think we need to move on to manual derivation (Trait implementation)

Here [is the conversion guideline](https://pyo3.rs/v0.16.2/conversions/tables.html) we should follow
for Python<->Rust conversion

that would be something like `IntoPyList` for vectors, and  `IntoPyDict` for all record types ?

README.md
```diff
@@ -90,6 +95,9 @@ select a SBAS for a given location on Earth.
 * `--flate2`  
 allow native parsing of .gz compressed RINEX files. Otherwise, user must uncompress manually the `.gz` extension first.
 
+* `--pyo3` (experimental)
+Add Python bindings via `PyO3`. To build the Python package, you must first install maturin and then build it with the pyo3 feature flag. For example, `maturin build -F pyo3`. Maturin will then build and place the resulting .whl file in `/target/wheels/`, after which you can install the package with `pip install rinex`.
```
gwbres, line 99: It would be awesome to be able to auto generate the wheel

rinex/src/epoch/flag.rs
gwbres: The basics is there, it is now working :) 
An example is attached and will get incremented.
Now we need to tackle the hardest parts 😄

README.md
larsnaesbye: So it gets included with the release, and later uploaded to pypi.org ?

gwbres: > So it gets included with the release

yes, like the application binaries

> and later uploaded to pypi.org ?

i did not think about going this far, but I don't foresee any reasons that would not be doable.

gwbres: Ok so [maturin-action](https://github.com/PyO3/maturin-action) does exist,
i just pushed a commit that integrates the basis. I simply reworked a basic example, we need to test it

rinex/src/python.rs
```diff
@@ -0,0 +1,26 @@
...
+};
+
+#[pymodule]
+fn rinex(_py: Python, m: &PyModule) -> PyResult<()> {
+    /*
+     * prelude module
+     */
+    m.add_class::<Epoch>()?;
+    m.add_class::<EpochFlag>()?;
+    /*
```
gwbres, line 18: we should design python modules here, following the crate architecture. This would allow:

```python
from rinex import Epoch
from observation import ObservationData
```

gwbres: In the very near future, I will only focus on solving the wheel generator: ideally fix the build on MacOS and unlock the packaging in a release

gwbres:
For the moment, I will only focus on solving the MacOS builder and unlock the wheel packager. 
We have the basis for most of the crate. 
We still need to figure out what needs to be done for Record structures (dictionaries, `Map<K, V>`) and we probably have to submit a PR to `PyO3` for the nested Enum case

larsnaesbye:
> and we probably have to submit a PR to `PyO3` for the nested Enum case.

Strange that they have not encountered this need before. But I agree.

rinex/pyproject.toml
```diff
@@ -0,0 +1,15 @@
+[build-system]
+requires = ["maturin>=0.14"]
+build-backend = "maturin"
+
+[project]
+name = "rinex"
+requires-python = ">=3.9"
```
gwbres, line 7: any particuliar reason for the `>=3.9` contraint ?

larsnaesbye: No, not other than that was the version I had installed when I wrote it.

gwbres: If you don't mind I decremented to v3.8, the wheels are auto generated for 3.8, 3.9 and 3.10 conveniently

larsnaesbye: That's perfectly fine. A lot of people still use 3.7 as far as I know.

rinex/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+build-backend = "maturin"
+
+[project]
+name = "rinex"
+requires-python = ">=3.8"
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: Apache Software License"
+    "License :: OSI Approved :: MIT License",
```
gwbres, line 12: the CI complains about a few missing "," here ;)

larsnaesbye: Ah yes - added two now

rinex/src/epoch/flag.rs
gwbres: > and we probably have to submit a PR to `PyO3` for the nested Enum case

we don't necessarily have to go this far. The immediate problem is the `pyclass` auto derivation does not support nested enums, that means this macro cannot dictate the code in this scenario.
We just have to to the job ourselves.
And it's probably a manual `IntoPyObject` implementation [like these ones](https://github.com/PyO3/pyo3/blob/249c0209fd41336b9e68bae14290e598fcad34ea/src/conversions/chrono.rs)

rinex/pyproject.toml
gwbres: thank you

gwbres:
Hello @larsnaesbye,

the obvious solution to bind complex enums, is to write a Python compliant version, for instant `PyConstellation` that would be possible to bind. But for me, this is a no-go because it would require to much rework of the entire library.

I have not found much information about the `ToPyObject` trait, there is plenty for `FromPyObject`, which we do not need. So I'm still not excluding a manual implementation would be doable. We could benefit asking the `PyO3` community.

Finally, I barely looked at the `PyO3` code, I still think providing the support for complex enums would be possible. But abviously introduces a lot of a latency until we can release our usage

larsnaesbye:
Looks like we need to rebase. I will look at squashing my commits soon.

gwbres:
Hello Lars,
align your fork/main and I'll take care of rebasing this PR ;)

larsnaesbye:
> Hello Lars,
> align your fork/main and I'll take care of rebasing this PR if it helps ;)

Good 🙂 , but what do you mean by aligning? I could probably just resolve conflicts using the graphical interface here on Github, I'm just unsure of `ephemeris.rs` - the others seem fairly easy.

larsnaesbye:
@gwbres I resolved the conflict - now on to solving the errors :-D

larsnaesbye:
Rebase done (in a roundabout way, I know...)

larsnaesbye:
Should this work start all over, with the new codebase, new PyO3 and maturin etc.? It's beginning to get a bit hairy....

gwbres:
> Should this work start all over, with the new codebase, new PyO3 and maturin etc.? It's beginning to get a bit hairy....

We only have a couple of useful commits on this branch, it should not take too long to start from scratch, and is probably easier.

Now I understand that we must provide alternative structures, for structures that cannot be bound right away.  
Cast in both ways must be provided

larsnaesbye:
Okay, in that case I will close the pull request, so we can start over when we have the time and energy :-)

larsnaesbye:
Closing to make way for a better PR later.

PR #54, opened 2022-11-28
switch to latest version of docs link

larsnaesbye:


gwbres (approved):

gwbres:
Thank you for your contribution 👍 
I initiated a general review of our "auto" docs, we should be close to the truth. 
Especially for the "cli" where I always provide the exact command line(s) that I use on my side

PR #55, opened 2022-11-29
Add task: automatic building of binaries

larsnaesbye:
Adding this because we might easier get feedback from GNSS professionals who are not Rust proficient if they can download a binary of 'rinex-cli'.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,14 +20,14 @@ jobs:
     steps:
       - uses: actions/checkout@master
       - name: Build dependencies 
-        run: sudo apt-get install libfontconfig1-dev
+        run: sudo apt-get update && sudo apt-get install libfontconfig1-dev
       - name: Compile and release
         uses: rust-build/rust-build.action@v1.4.0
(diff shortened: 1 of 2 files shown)


gwbres:
@larsnaesbye 

are you aware of a method to trigger this job so we can debug it without releasing a tag?
I pushed some corrections that the previous release generated, but we're still unusure this will properly work until next time

larsnaesbye:
@gwbres I don't know an easy way to do it other than what I did - try it out on a separate demo repository.
There is a check box "This is a pre-release" that marks the release as not ready for production, and releases can be deleted afterwards.

gwbres:
> @gwbres I don't know an easy way to do it other than what I did - try it out on a separate demo repository. There is a check box "This is a pre-release" that marks the release as not ready for production, and releases can be deleted afterwards.

Ok I'll just push 1 or 2 pre releases until we stabilize this feature. We should be very close, the most problematic was my introduction of the linux non musl, which is clearly stated as not supported by this tool (official page).

PR #72, opened 2022-12-18
ignore macOS .DS_Store files

larsnaesbye:
These turn up all the time on macOS, so we should ignore them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Cargo.lock
 *.jpg
 *.swp
 *.swo
+.DS_Store


gwbres:
Is this a subfolder ? otherwise you need a *.DS wildcard to hide all hidden files that start with that pattern

larsnaesbye:
Here on my machine, simply adding it like I did here removed the three instances I used to have in my repo.


Before:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.DS_Store
	rinex/.DS_Store
	rinex/src/.DS_Store
```

After:
```
nothing to commit, working tree clean
```

gwbres (approved):

PR #87, opened 2023-05-17
Replace deprecated INFINITY

larsnaesbye:


larsnaesbye:
It could also instead be replace with NEG_INFINITY. Not much of a difference :-)

gwbres (approved):

PR #123, opened 2023-08-08
fix feature definition

gwbres:
These methods should be implemented only when the "processing" feature is enabled.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -26,11 +26,11 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y libudev-dev
     - name: Build (Default)
-      run: cargo build
-    - name: Build (All features)
-      run: cargo build --all-features
+      run: cargo clean && cargo build
     - name: Test (Default)
       run: cargo test --features tests -- --nocapture
+    - name: Build (All features)
+      run: cargo clean && cargo build --all-features
     - name: Test (All features)
       run: cargo test --all-features -- --nocapture
     - name: Coding style
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -26,11 +26,11 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y libudev-dev
     - name: Build (Default)
-      run: cargo clean && cargo build
+      run: cargo build
     - name: Test (Default)
       run: cargo test --features tests -- --nocapture
     - name: Build (All features)
-      run: cargo clean && cargo build --all-features
+      run: cargo build --all-features
     - name: Test (All features)
       run: cargo test --all-features -- --nocapture
     - name: Coding style
--- a/rinex/src/ionex/record.rs
+++ b/rinex/src/ionex/record.rs
@@ -1,12 +1,4 @@
-use crate::{
-    algorithm::{Filter, Interpolate, Mask, MaskFilter, MaskOperand, Preprocessing, TargetItem},
-    gnss_time::GnssTime,
-    merge,
-    merge::Merge,
-    prelude::*,
-    split,
-    split::Split,
-};
+use crate::{gnss_time::GnssTime, merge, merge::Merge, prelude::*, split, split::Split};
 
 use super::{grid, GridLinspace};
 
@@ -479,6 +471,10 @@ impl GnssTime for Record {
     }
 }
 
+#[cfg(feature = "processing")]
+use crate::preprocessing::*;
+
+#[cfg(feature = "processing")]
 impl Mask for Record {
     fn mask(&self, mask: MaskFilter) -> Self {
         let mut s = self.clone();
@@ -515,6 +511,7 @@ impl Mask for Record {
     }
 }
 
+#[cfg(feature = "processing")]
 impl Preprocessing for Record {
     fn filter(&self, f: Filter) -> Self {
         let mut s = self.clone();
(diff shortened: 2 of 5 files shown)


gwbres:
Why is this not picked up by the CI script ? 

We build the library twice: 

- cargo build  --default : no features enabled, this should have been picked up
- cargo build --all-features

gwbres:
So far, my explanation is, it did pass if we have some `build --all-features` residues, 
a  `cargo clean` is probably required prior anything

larsnaesbye (approved):

.github/workflows/rust.yml
```diff
@@ -26,11 +26,11 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y libudev-dev
     - name: Build (Default)
-      run: cargo build
-    - name: Build (All features)
-      run: cargo build --all-features
+      run: cargo clean && cargo build
```
lnicola, line 29: I don't think this is actually necessary, and I can't reproduce any build failure.

lnicola:
```
$ git fetch origin && git checkout main
Already on 'main'
Your branch is up to date with 'origin/main'.
$ cargo clean && cargo build -q
warning: unused variable: `clk_data`
   --> rinex/src/qc/analysis/obs.rs:309:17
    |
309 |             let clk_data = rec.filter(mask);
    |                 ^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_clk_data`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: variable does not need to be mutable
   --> rinex/src/qc/analysis/obs.rs:302:13
    |
302 |         let mut clock_drift: Option<f64> = None;
    |             ----^^^^^^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: field `clock_diff` is never read
  --> rinex/src/hatanaka/compressor.rs:43:5
   |
23 | pub struct Compressor {
   |            ---------- field in this struct
...
43 |     clock_diff: NumDiff,
   |     ^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field `clock_drift` is never read
   --> rinex/src/qc/analysis/obs.rs:280:5
    |
255 | pub struct QcObsAnalysis {
    |            ------------- field in this struct
...
280 |     clock_drift: Option<f64>,
    |     ^^^^^^^^^^^
    |
    = note: `QcObsAnalysis` has derived impls for the traits `Clone` and `Debug`, but these are intentionally ignored during dead code analysis

warning: use of deprecated method `chrono::NaiveDate::and_hms`: use `and_hms_opt()` instead
  --> sinex/src/datetime.rs:20:11
   |
20 |     Ok(dt.and_hms(h as u32, m as u32, s as u32))
   |           ^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

$
```

gwbres: So it passed.. this is confusing.

When building with no features, "processing" feature is disabled.
But `src/ionex/record.rs` tries to import processing features whether it is enabled or not

lnicola: Ah, I get the failure with `cargo build --all-targets`.

gwbres: Something happens when you invoke cargo twice, with different options.

lnicola: It's because `rinex-cli` enables the `processing` feature. `cargo build -p rinex` doesn't (and neither does `cargo package`, I suppose), so you get the compilation errors. I'd use `--all-targets`, but it currently doesn't work because of some code that's now missing.

lnicola (changes requested):
Looks okay except for the workflow changes. I'd try to enable `--all-features` there instead. The `cargo clean` there is misleading.

Medium-term, it might be a good idea to move anything gated by `processing` into a `processing` module, e.g. `src/processing/record.rs`, which should make feature gates easier to manage.

.github/workflows/rust.yml
gwbres: @lnicola 

>  I'd use --all-targets

The problem with this, is it seems to end up the tests without any flags.
And we need at least `--feature tests` to be passed when running these methods,
to include the test methods. I have grown tiered to write and rewrite the same thing all the time, and it makes the test infra much more efficient to share methods.

Unfortunately I have not found any clean methods to include shared methods into different "tests".  
The only way is to put it in the library directly.

lnicola: Okay, plan B is to run `cargo build -p` on each crate :-(.

lnicola: Can you gate `test_toolkit` on `cfg(test)` instead?

gwbres:
> Looks okay except for the workflow changes. I'd try to enable `--all-features` there instead. The `cargo clean` there is misleading.
> 
> Medium-term, it might be a good idea to move anything gated by `processing` into a `processing` module, e.g. `src/processing/record.rs`, which should make feature gates easier to manage.

I reverted that script

.github/workflows/rust.yml
gwbres: that's a good idea. I'll give it a try.

How can I make the "requested changed" as solved ?

gwbres: > Can you gate `test_toolkit` on `cfg(test)` instead?

that is already the case, see the first few lines, for rinex/src/lib.rs

lnicola: I don't think I see it, can you link? I meant cfg(test), not the tests feature.

lnicola (approved):

PR #125, opened 2023-08-09
add documentation to cargo.toml

gwbres:


gwbres:
@larsnaesbye, @lnicola 

For some reason, the crate has been detached from the doc.rs portal, and the latest documented version is stil 0.10.x.  
Is that because we don't have an "documentation" field right here? hopefully that will fix it

lnicola:
There's no need to do this, it will show up.

gwbres:
> There's no need to do this, it will show up.

it did! :)

PR #127, opened 2023-08-11
update minimum rustc requirement

gwbres:
The use of "indexmap"/"hashbrown" has increased the minimumt rustc required. 
I think it is tied to IONEX/maps operations.
The ionex infra will get finalized and aligned to latest work, so that probably will not last for long.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ RINEX
 [![crates.io](https://docs.rs/rinex/badge.svg)](https://docs.rs/rinex/)
 [![crates.io](https://img.shields.io/crates/d/rinex.svg)](https://crates.io/crates/rinex)
 
-[![minimum rustc: 1.61](https://img.shields.io/badge/minimum%20rustc-1.61-blue?logo=rust)](https://www.whatrustisit.com)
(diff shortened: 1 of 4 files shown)


larsnaesbye (approved):

gwbres:
> You can put it in the manifest so it gets checked automatically.

We already had 

```toml
[package]
rust-version = "1.61"
```

that I forgot to increment. Also `rinex-cli` needs it, since it uses absolutly all features from `rinex`.   
But apparently we don't have an automated verification, how do you do that ?

lnicola (approved):

lnicola:
FWIW, this can also be set in the workspace manifest: https://github.com/rust-lang/rust-analyzer/blob/master/Cargo.toml#L7, https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide-assists/Cargo.toml#L9.

gwbres:
> FWIW, this can also be set in the workspace manifest: https://github.com/rust-lang/rust-analyzer/blob/master/Cargo.toml#L7, https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide-assists/Cargo.toml#L9.

Ok, we can rework the workspace definition in a separate PR.
I'm not going to touch the CI script in this PR also, the current "rust stable toolchain" seems to satisfy v1.64, and checking the minimum version will require a dedicated job

Issue #117, opened 2023-08-05
Cli: fix the skyplot view


gwbres:
When at least one NAV file is provided, a skyplot view is feasible.

The current view is not working correctly

PR #130, opened 2023-08-24
fix kepler solver and skyplot view

gwbres:
Fix Kepler equations solver and visualization method

- #117

Although previous computations were 90% correct, we did not operate the calculators correctly, both in
- [x] `Rinex::sv_position` 
- [x] and `Rinex::sv_elev_azim`

mainly due to wrong time scales being used and time conversion issues.

Several unit tests are introduced and a python helper script helps generate test data against third party tools.

The skyplot viewer has been corrected, with Rho=elev, and Theta=azim visualization.


gwbres:
Fixed the viewer :) 
skyplot

lnicola (approved):

PR #132, opened 2023-08-24
glonass and sbas positioning

gwbres:
Skyplot view for GLO and GEO constellations.
Satellite positions are broadcasted in ECEF expressed in km : data is already available.

### DIFF
--- a/rinex/src/navigation/ephemeris.rs
+++ b/rinex/src/navigation/ephemeris.rs
@@ -364,12 +364,18 @@ impl Ephemeris {
      * Either by solving Kepler equations, or directly if such data is available.
      */
     pub(crate) fn sv_position(&self, sv: &Sv, epoch: Epoch) -> Option<(f64, f64, f64)> {
-        if let Some(pos_x) = self.get_orbit_f64("satPosX") {
-            if let Some(pos_y) = self.get_orbit_f64("satPosY") {
-                if let Some(pos_z) = self.get_orbit_f64("satPosZ") {
-                    //TODO PZ90=>ECEF transform should be applied, in case of GLONASS
-                    //     need to check for GEO/SBAS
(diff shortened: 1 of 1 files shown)


gwbres:
See this figure, where I plot GPS against GLO to make sure the scale is consistent.

image

rinex/src/navigation/ephemeris.rs
```diff
@@ -364,12 +364,18 @@ impl Ephemeris {
      * Either by solving Kepler equations, or directly if such data is available.
      */
     pub(crate) fn sv_position(&self, sv: &Sv, epoch: Epoch) -> Option<(f64, f64, f64)> {
-        if let Some(pos_x) = self.get_orbit_f64("satPosX") {
-            if let Some(pos_y) = self.get_orbit_f64("satPosY") {
-                if let Some(pos_z) = self.get_orbit_f64("satPosZ") {
-                    //TODO PZ90=>ECEF transform should be applied, in case of GLONASS
-                    //     need to check for GEO/SBAS
-                    return Some((pos_x, pos_y, pos_z));
+        if let Some(pos_x_km) = self.get_orbit_f64("satPosX") {
```
lnicola, line 367: This could be written as an separate function / lambda to use `?`, but it's fine like this too.

lnicola (approved):

rinex/src/navigation/ephemeris.rs
gwbres: The thing here, it should not "return" in case those fields are missing. When those fields are missing, we should execute `kepler2ecef` that comes right after this if/else

lnicola: Yeah, that's what I meant by moving the code to a new function:

```rust
self.sat_position().or_else(|| self.kepler2ecef(sv, epoch))
```

Issue #133, opened 2023-08-25
NAV V4 Ephemeris support


gwbres:
The NAV frame descriptor (build script) needs to be augmented, as data fields depend on the `NavMsgType` since NAV V4.

Tied to 

- [ ] #78 
- [ ] #80

Issue #80, opened 2023-04-15
Ephemeris health fields


ljbade:
I ended up having to modify the [`orbits.json`](https://github.com/gwbres/rinex/blob/main/rinex/db/NAV/orbits.json) file as I was having problems with the decoding of the health bits. I think this was mainly due to the mix of LNAV and CNAV types in there at the moment. The different ephemeris types tend to have quite different definitions of how they encode the health information.

gwbres:
@ljbade 
the orbits descriptor should be more complex, to fully describe modern navigation frames

larsnaesbye:
@ljbade Would you like to share the changes in a pull request?

ljbade:
Sure I can put up the changes I made, there not pretty but was enough for my immediate need. Shouldn't be too hard to polish it up though.

gwbres:
Hello @ljbade,

I started working on this issue [in this branch](https://github.com/georust/rinex/compare/main...nav-health).
It will take a little time for this to conclude, feel free to join the effort.

The documentation are rather unclear on how to correctly interpret bitfields, the only one I found straightforward so far is Galileo. For example, I'm having a hardtime relate the pseudo specifications in the RINEX4.pdf and the QZSS specs.

gwbres:
> I agree the RINEX document doesn't always line up clearly with the ICDs. Hopefully I will get some time to check your branch later in the week.

Awesome. Yeah.. it's hard to consider those as real _specifications_: 

![image](https://github.com/georust/rinex/assets/7270467/04e5f7a2-6b60-43e1-953c-36ca41d49cf7)

bit[17]..[22] I presume represent the 6 bit mentionned in the QZSS specs, dunno what `w 3 sf 1" means.

I also don't know how to interpret GPS CNAV1/CNAV2, while GPS LNAV is correctly documented

![image](https://github.com/georust/rinex/assets/7270467/a0f2791d-ff65-43bd-bdab-3809018f7e7c)

ljbade:
So here is Word 3 of Subframe 1 of the "LNAV" data for QZSS taken from IS-QZSS-PNT-005 section 4.1.2.3 (pg 46):
![image](https://github.com/georust/rinex/assets/1215089/ca56fef2-9b1a-4425-bf33-0aca72d26daf)

Basically they are talking about the "SV HEALTH" 6 bits field.

The next page gives the interpretation of these bits:
[...]

Issue #78, opened 2023-04-14
Support CNAV ephemeris types


larsnaesbye:
### Discussed in https://github.com/gwbres/rinex/discussions/76

<div type='discussions-op-text'>

<sup>Originally posted by **ljbade** March 17, 2023</sup>
With RINEX 4 it is now possible to get broadcast ephemerides for the newer signals such as GPS CNAV or BDS CNAV. However it seems the current implementation only recognises the ephemeris types from RINEX 3 such as LNAV.

I am interested in computing the position difference between the different ephemeris types for the same satellite so it would be useful to add support for:
- GPS CNAV and CNAV-2
- QZSS CNAV and CNAV-2
- BDS CNAV-1, CNAV-2 and CNAV-3

The recent IGS broadcast navigation files have examples of most of these new formats. An example is https://cddis.nasa.gov/archive/gnss/data/daily/2023/brdc/BRD400DLR_S_20230750000_01D_MN.rnx.gz

The IGS file doesn't seem to yet have an example of GPS CNAV-2 yet, which is from the GPS L1C signal.</div>

ljbade:
A few more notes based on some of the hacky fixes I made to get CNAV stuff working good enough for what I needed.

I think the JSON file `orbits.json` should be expanded for RINEX 4 to allow an ephemeris type map where the keys are the RINEX names (e.g. "LNAV" "CNAV", "CNV2"). Then as the reader is parsing the file it can use the correct field name map for each type.

I noticed that for some reason there is CNAV formats in the `orbits.json` file but they are marked as version 4, and the legacy ones are marked as version 3. So my hacky fix was just to switch the RINEX version number if I was wanting the CNAV data.

ljbade:
Putting this here as a reference, for BDS CNAV this is the field names I ended up using:
```
            {
                "revision": {
                    "major": 4
                },
                "content": {
                    "adot": "f64",
                    "crs": "f64",
                    "deltaN": "f64",
                    "m0": "f64",
                    "cuc": "f64",
                    "e": "f64",
                    "cus": "f64",
                    "sqrta": "f64",
                    "toe": "f64",
                    "cic": "f64",
                    "omega0": "f64",
                    "cis": "f64",
                    "i0": "f64",
                    "crc": "f64",
                    "omega": "f64",
                    "omegaDot": "f64",
                    "idot": "f64",
                    "deltaNDot": "f64",
                    "satType": "f64",
                    "t_op": "f64",
[...]

PR #135, opened 2023-08-25
NAV V4: RINEX V4 navigation support

gwbres:
Preliminary work to support CNAV, CNAV-2 and other modern Ephemeris frames.

----

Answers 

- [x] #133

preliminary work to provide an answer to 

- #78
- and #80

To fully support NAV-V4, we need a slightly more complex Navigation Frame database, 
because the frames content now depends on `NavMsgType` also.
Therefore, this field needs to be introduced in the database, and taken care of.

This branch only introduces the foundation and we still only truly support `NavMsgType::LNAV` only.  
Modern frames will be introduced in following contributions.

  - Modified (and simplified) the JSON description to describe the `NavMsgType`. 

This modification also impacts the Rinex filter designer, because filtering on Nav Orbit Fields is one possible type of filter.

The `build.rs` script is slightly improved:

  - Don't need to use '&str' fields everywhere! 
We can use Rinex types directly (like `Constellation` or `NavMsgType`).
This saves the `&str` interpretation and cast to these structures, when we actually browse the NAV RINEX standards

  -  Improved the `closest_nav_standards` utility, which browse our DataBase when we need to identify an navigation. It previously returned the closest revision. We still had to actually grab its content. Now we return the entire Frame Description, which makes it easier when parsing.

-  Add more thorough `closest_nav_standards` tests

### DIFF
--- a/rinex/src/navigation/ephemeris.rs
+++ b/rinex/src/navigation/ephemeris.rs
@@ -119,7 +119,7 @@ impl Ephemeris {
             .insert(field.to_string(), OrbitItem::from(value));
     }
     /// Retrieve week counter, if such data exists.
-    pub fn get_weeks(&self) -> Option<u32> {
+    pub fn get_week(&self) -> Option<u32> {
         self.orbits.get("week").and_then(|field| field.as_u32())
     }
     /*
@@ -319,7 +319,7 @@ impl Ephemeris {
         let kepler = self.kepler()?;
         let perturbations = self.perturbations()?;
 
-        let weeks = self.get_weeks()?;
+        let weeks = self.get_week()?;
         let t0 = GPST_REF_EPOCH + Duration::from_days((weeks * 7).into());
         let toe = t0 + Duration::from_seconds(kepler.toe as f64);
         let t_k = (t_sv - toe).to_seconds();
@@ -467,7 +467,9 @@ fn parse_orbits(
             if content.trim().len() == 0 {
                 // omitted field
                 key_index += 1;
-                nb_missing -= 1;
+                if nb_missing > 0 {
+                    nb_missing -= 1;
+                }
                 line = rem.clone();
                 continue;
             }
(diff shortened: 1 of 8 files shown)


rinex/src/navigation/ephemeris.rs
```diff
@@ -245,7 +242,7 @@ impl Ephemeris {
         s
     }
     /// Creates new Ephemeris with given week counter
-    pub fn with_weeks(&self, week: u32, constellation: Constellation) -> Self {
+    pub fn with_week(&self, week: u32) -> Self {
```
gwbres, line 245: Week counter is not tied to a specific constellation, remove unused argument

rinex/src/navigation/orbits.rs
```diff
@@ -230,175 +226,191 @@ impl OrbitItem {
...
+        let items: Vec<_> = database
             .iter()
-            .map(|r| {
-                (
-                    u8::from_str_radix(r.major, 10).unwrap(), // to later build object
-                    i8::from_str_radix(r.major, 10).unwrap() - desired_rev.major as i8, // for filter op
-                    u8::from_str_radix(r.minor, 10).unwrap(), // to later build object
-                    i8::from_str_radix(r.minor, 10).unwrap() - desired_rev.minor as i8, // for filter op
-                )
+            .filter_map(|item| {
```
lnicola, line 253: This could be `.filter(|item| item.constellation == constellation && item.msg == msg && item.version == Version::new(major, minor)`, right?

lnicola (approved):
Mostly rubber-stamping, but seems okay.

larsnaesbye (approved):

gwbres:
Hum.. I picked up some errors when parsing Old NAV(V2) (only Glonass and GPS exist in those cases), we could use more tests of those cases. Publishing them in a bit

gwbres:
> Hum.. I picked up some regressions, but it is unclear why this contribution is causing them. Either way, we could use more tests about NAV parsing. I will introduce more of them and will inquire

false alarm, it's just this branch was not aligned with the latest `main` branch. Dunno why github marks it as "ready to merge" then.. Anyways.. the two more tests I introduced won't hurt either

rinex/src/navigation/orbits.rs
gwbres: correct! applied

lnicola:
> Dunno why github marks it as "ready to merge" then

See also https://graydon.livejournal.com/186550.html :-).

gwbres:
Hello,

I see that in his repo rust-lang/bors, the library tests are built in the
crate, not in `tests/` folder.
But only included on `cfg[test]`. This would solve the problem to include
methods that are shared amongst all tests

Guillaume W. Bres
Software engineer
***@***.***>


Le sam. 26 août 2023 à 11:03, Laurențiu Nicola ***@***.***> a
écrit :

lnicola:
I didn't mean the structure of the bors specifically, but rather the idea of merging branches and running tests again. An old branch might have passed the tests when it was open, but it might no longer do once it's merged, because of other changes made in the meanwhile, even if there were no merge conflicts.

Anyway, it's only incidentally relevant here.

PR #139, opened 2023-08-26
Release v0.12.0

gwbres:


larsnaesbye:
Looking good! Should we get the crate up on https://georust.org too?

rinex-cli/src/plot/record/navigation.rs
```diff
@@ -43,6 +43,7 @@ pub fn plot_navigation(rinex: &Rinex, plot_ctx: &mut PlotContext) {
             epochs.clone(),
             sv_clock,
         )
+        .web_gl_mode(true)
```
lnicola, line 46: I wonder what this does in a CLI app :⁠-⁠)

lnicola (approved):

PR #141, opened 2023-08-27
Introduce observation data iterators

gwbres:
Introduce new OBS RINEX iteration methods, similar to other methods recently developed.

* carrier_phase() : iterator over raw phase data
* pseudo_range(): iterator over pseudo range estimates
* ssi(): iterator over estimated signal strength
* doppler(): iterator over doppler shifts

### DIFF
--- a/rinex/src/lib.rs
+++ b/rinex/src/lib.rs
@@ -1874,9 +1874,8 @@ impl Rinex {
         }))
     }
     /// Returns an iterator over receiver clock offsets, expressed in seconds.
-    /// Such information is kind of rare (modern receivers?),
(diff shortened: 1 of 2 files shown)


lnicola (approved):
These aren't from a trait, so they could return `impl Iterator` directly, but it's fine anyway.

larsnaesbye (approved):

gwbres:
> These aren't from a trait, so they could return `impl Iterator` directly, but it's fine anyway.

That is correct, but that could also apply to all other existing Iterators.
Let's simplify their signatures in a dedicated PR

PR #143, opened 2023-08-28
simplify Iterators definition

gwbres:
Simplify returned type, when we return an `Iterator` : it never requires it to be `Box`'ed

- Answers #142

### DIFF
--- a/rinex/src/lib.rs
+++ b/rinex/src/lib.rs
@@ -1334,30 +1334,28 @@ impl Rinex {
     ///     "sampling_histogram failed"
     /// );
     /// ```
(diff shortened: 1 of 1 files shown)


gwbres:
@lnicola,

there are a few spots where I can't see to remove the `Box`, why is that ?

For example, this does not compile
image

gwbres:
> The reason is that `impl Iterator` is always a single type, but the code in the screenshot would return 5 different types

the returned type is the same, in the sense they're all Iterators over an `Epoch`, but the iterator source is different that is true

lnicola:
The closures are also different types, even if they have the same arguments and return value (and I'm not sure these ones do).

PR #149, opened 2023-08-29
Fix RINEX v4 mistakes in orbits.json

ljbade:


rinex/db/NAV/orbits.json
```diff
@@ -295,9 +299,9 @@
             "status": "gloStatus",
             "L1L2GrpDelay": "f64",
             "urai": "f64",
-			"almanacHealth": "health"
+            "almanacHealth": "health"
```
ljbade, line 302: Seems VSCode auto formatted this

rinex/db/NAV/orbits.json
```diff
@@ -680,20 +688,22 @@
...
             "sisa1_oe": "f64",
             "sisa1_ocb": "f64",
             "sisa1_oc1": "f64",
             "sisa1_oc2": "f64",
-            "isc_b1cd": "f64",
-            "tgd_b1cp": "f64",
-            "tgd_b2ap": "f64",
+            "iscB1Cd": "f64",
+            "spare1": "xxx",
+            "tgdB1Cp": "f64",
```
ljbade, line 698: I changed the naming style of the BDS stuff to be similar to GPS/QZSS, one advantage of camel case is that I can correctly mimic the capital letters in the signal name which has special meaning and matches BDS ICDs

rinex/db/NAV/orbits.json
```diff
@@ -680,20 +688,22 @@
...
+            "tgdB2ap": "f64",
             "sismai": "f64",
             "health": "f64",
-            "b1c": "f64",
+            "integrityB1C": "f64",
```
ljbade, line 702: I ended up putting `integrity` first, to match `isc` and `tgd` - means I can use the camel case in the signal name

If you prefer to keep the naming from the RINEX specification then I can put integrity last but then would have to make lower case `b`

gwbres: no that's alright, IMO it's better to have consistent names at the expense of grammar correctness

rinex/db/NAV/orbits.json
gwbres: good idea

PR #156, opened 2023-09-01
Improve Qc context definition

gwbres:
Define QcPrimaryData to define primary file.
Define abstract QcExtraData to define abstract context augmentations. (either Rinex or SP3) 
We can now form a context blob that results from loading several files of the same kind.
This allows analyzing a context that can span more than 24h for example.


rinex/src/qc/context.rs
```diff
@@ -1,61 +1,93 @@
...
+    pub fn data(&self) -> &T {
+        &self.data
+    }
+    /// Returns mutable reference to Inner Data
+    pub fn data_mut(&mut self) -> &mut T {
+        &mut self.data
+    }
+    /// Returns list of files that created this context
+    pub fn paths(&self) -> Vec<PathBuf> {
+        self.paths.clone()
```
gwbres, line 37: Hello @lnicola,

any recommendations to avoid this `.clone()` ? due to the fact we're retuning the List of PathBufs.

Not that this matters so much, this is only used in two locations
- when building the actual Context provided by the entire command line.
We need this to append a new path for example if this is the second file of this kind
- when reporting the final anlysis, in the report header we notify the file locations, and this is invoked once

gwbres:
Next Context report : 

image

rinex-cli/src/cli.rs
```diff
@@ -24,7 +24,7 @@ impl Cli {
                     .arg(Arg::new("filepath")
                         .short('f')
                         .long("fp")
-                        .value_name("FILE")
+                        .value_name("[FILE]")
```
lnicola, line 27: Not sure if this is actually an improvement.

rinex-cli/src/main.rs
```diff
@@ -74,77 +69,97 @@ pub fn create_workspace(path: PathBuf) {
...
-            if let Some(path) = cli.atx_path() {
-                if let Ok(ctx) = QcInputData::new(path) {
-                    if ctx.rinex.is_antex() {
-                        trace!("atx data \"{}\"", path);
-                        Some(ctx)
-                    } else {
-                        error!("--atx should be valid antex data");
-                        None
-                    }
+                if ctx.paths().len() == 0 {
```
lnicola, line 87:
```suggestion
                if ctx.paths().is_empty() {
```

rinex-cli/src/main.rs
```diff
@@ -74,77 +69,97 @@ pub fn create_workspace(path: PathBuf) {
...
+                    // first file passed by user
+                    ctx = QcExtraData {
+                        paths: vec![Path::new(path).to_path_buf()],
+                        data: new.clone(),
```
lnicola, line 91: Is the `clone` needed here?

rinex-cli/src/main.rs
```diff
@@ -74,77 +69,97 @@ pub fn create_workspace(path: PathBuf) {
...
+                    };
+                    trace!("{} file \"{}\"", rtype, path);
                 } else {
-                    error!("failed to parse atx data \"{}\"", path);
-                    None
+                    match ctx.data_mut().merge_mut(&new) {
+                        Ok(_) => {
+                            trace!("{} file \"{}\"", rtype, path);
+                            ctx.paths.push(Path::new(path).to_path_buf());
```
lnicola, line 98: `PathBuf::from`

rinex-cli/src/main.rs
```diff
@@ -74,77 +69,97 @@ pub fn create_workspace(path: PathBuf) {
...
+}
+/*
+ * Macro to build QcExtraInput of type SP3 specifically
+ */
+fn build_extra_sp3_data(paths: Option<ValuesRef<'_, String>>) -> Option<QcExtraData<SP3>> {
+    if let Some(paths) = paths {
+        let mut ctx = QcExtraData::<SP3>::default();
+        for path in paths {
+            if let Ok(new) = SP3::from_file(path) {
+                if ctx.paths().len() == 0 {
```
lnicola, line 121: `is_empty`

rinex/src/qc/context.rs
lnicola: Easiest is to return a `&[PathBuf]`.

rinex/src/qc/context.rs
```diff
@@ -1,61 +1,93 @@
...
+            nav.paths()
+        } else {
+            vec![]
+        }
     }
     /// Returns reference to navigation data specifically
     pub fn navigation_data(&self) -> Option<&Rinex> {
         if let Some(ref nav) = self.nav {
-            Some(&nav.rinex)
+            Some(&nav.data)
```
lnicola, line 82: `.map` might be more idiomatic, but doesn't matter.

lnicola (approved):

rinex-cli/src/cli.rs
gwbres: Solved, by improving the helper notes on all of these

rinex-cli/src/main.rs
gwbres: `PathBuf::from` can't apply here, because `path` does not have a known size at compilation time ([u8]). Path comes from 

```rust
for path in clap::ValuesRef<'_, String> 
```

and clap::ValuesRef is the only way I know, at the moment, to have a command line argument pass more than one value

rinex/src/qc/context.rs
gwbres: can't seem to find a more idiomatic way, that would not involve a Copy

Rinex does not implement copy. Currently it proposes something like this :

```rust 
pub fn navigation_data(&self) -> Option<&Rinex> {
      self.nav 
           .clone() // want to avoid that
           .map(|nav| nav.data)
           .as_ref()
}
```

rinex-cli/src/main.rs
gwbres: the cli.rs structure should cast to Strings and not simply return ValuesRef then

rinex-cli/src/main.rs
gwbres: That is a good catch actually and I probably need to discuss this with you.
Moving to discord ?

lnicola: Can't right now, feel free to merge as-is and I'll try to look at it in a couple of days.

gwbres: ;) I never mean "right now", I'll just leave you a message, feel free to anwser any time

rinex/src/qc/context.rs
gwbres: Applied, it works. Not used to that form but it's simple

PR #160, opened 2023-09-07
Move qc features into dedicated library

gwbres:
To continue the efforts to simplify and reduce the `rinex/` crate code quantity, it makes a lot of sense to move the QC capacities into a dedicated library.

- `rinex/` is the parser and file production library and should contain just that. 
It contains high level extraction method, it can also contain specific estimators, like `Rinex.phase()` or `Rinex.dcb()` that is fine.
But actual RINEX analysis should not take place in that library
- `rinex-qc` is a library that provides structures to analyze a `Rinex` or a pool of `Rinex` structures.
In the proposed form, it is solely the previous rinex::qc module moved into that library.
- `rinex-cli` uses both

We use traits to make `QcReports` synthesis easy.   
Traits can only be implemented on structures that are defined in that very library.  
`rinex` and `rinex-qc` both need to implement the reporting traits. To solve that dual dependency, I propose to have a small `qc-traits` library that both can import. The library in the proposed form, only contains the `HtmlReport` trait, which is our only way to render a Qc report at the moment


larsnaesbye (approved):
Looks good to me, and it makes it more clear with the new structure.

PR #161, opened 2023-09-07
only rinex-qc and cli need SP3 dependency

gwbres:
- [x] Only rinex-qc and rinex-cli need the SP3 library actually. It is no longer needed by the parser,
since we split the QC into a separate library
- [x] Improve general test environment. From time to time, we're facing false negative (false test failures), because tests get executed in a multi threaded environment and we use basic and similar names like "test.crx" or "tmp.rnx" in separate methods.
With this simple random name generator, we make sure each tasklet creates a unique file.  
We also made sure, in a previous contribution, that all temporary generated files get removed at the end of the test execution.

### DIFF
--- a/rinex/src/tests/production.rs
+++ b/rinex/src/tests/production.rs
@@ -1,24 +1,21 @@
 #[cfg(test)]
 mod test {
-    use crate::tests::toolkit::compare_with_panic;
+    use crate::tests::toolkit::{compare_with_panic, random_name};
     use crate::*;
     fn testbench(path: &str) {
         // parse this file
         let rnx = Rinex::from_file(path).unwrap(); // already tested elsewhere
-        let copy_path = path.to_owned() + "-copy";
-        assert_eq!(rnx.to_file(&copy_path).is_ok(), true); // test writer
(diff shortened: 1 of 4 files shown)


larsnaesbye (approved):

rinex/src/tests/production.rs
```diff
@@ -1,24 +1,21 @@
...
-    use crate::tests::toolkit::compare_with_panic;
+    use crate::tests::toolkit::{compare_with_panic, random_name};
     use crate::*;
     fn testbench(path: &str) {
         // parse this file
         let rnx = Rinex::from_file(path).unwrap(); // already tested elsewhere
-        let copy_path = path.to_owned() + "-copy";
-        assert_eq!(rnx.to_file(&copy_path).is_ok(), true); // test writer
-        let copy = Rinex::from_file(&copy_path);
+        let tmp_path = format!("test-{}.rnx", random_name(5));
```
lnicola, line 8: In `gdal` we use `tempfile` and something like https://github.com/georust/gdal/blob/c7f121c01748e92a271d960e7e48415c186a6c13/src/test_utils.rs#L11 for this.

Issue #81, opened 2023-04-15
Decoded Epoch type often has wrong timescale


ljbade:
The hifitime `Epoch` struct supports different timescales but the decoder often initialises them with the wrong types.

For reference here is how each type of RINEX file should set the timescale:

- clocks -> UTC
- meteo -> UTC
- observation
  -  use the value from the `TIME OF FIRST OBS` header line
  - mandatory header for RINEX 3+, but in RINEX 2 it is only requires for mixed GPS/GLONASS files
  - so if not present default to GPS for pure GPS file, and GLO for pure GLONASS file
- navigation (EOP/EPH/STO/ION) records
  - for RINEX 3+ the value is whatever constellation the satellite is in, so for each record look at the satellite ID, and use that
  - e.g. if the satellite ID is `G01` use GPS timescale, `E01` use Galileo etc
  - for RINEX 2 the value is the constellation type of the file (since RINEX 2 uses a different file for GPS or GLONASS)

ljbade:
While working on this I also found a bug in hifitime relates to the GNSS time scales - https://github.com/nyx-space/hifitime/issues/209 - so be sure to the dependency once thats fixes

Also there isn't support for IRNSS/QZSS yet in hifitime - https://github.com/nyx-space/hifitime/issues/210 - so might need to fall back to something else, for QZSS I would use GPS in the meantime. Seems like the decoder doesn't support IRNSS so not an immediate issue there.

gwbres:
Hello @ljbade,

I am not surprized by incorrect behaviors, due to the many possible cases, that you thoroughly described.
If you feel like correcting these initial conditions, feel free to submit a PR I'll be happy to merge it

ljbade:
OK, I will work on a small PR based on the changes I made.

ljbade:
Started on a PR in https://github.com/ljbade/rinex/tree/time-scale

I still need to check that it handles the different cases and add tests then I'll make a PR.

ljbade:
I have got stuck fixing up the epoch formatting functions to match the parse functions. I can't figure out how to get the Gregorian format from a `hifitime::Epoch` in some given time scale. There is only a function for returning the Gregorian values in UTC or TAI, and the `Epoch::compute_gregorian` function is private.

Basically I want to ensure the output RINEX formatted observation times are in the correct time scale for the type of file (e.g. GPS, Galileo, etc) to match the rules for parsing observation times.

gwbres:
[...]

PR #162, opened 2023-09-08
fix time scale interpretation

gwbres:
This branch solves the following issue

- [ ] #81 

Previous version of RINEX would identify most Epochs as UTC Epoch.   
This is incorrect, as per RINEX specifications. Refer to the related issue for more information.

This branch proposes a fix that solves the Navigation and Observation Data cases which are the most complex.  

To solve that, we use Epoch::from_str("YYYY-MM-DDTHH:MM:SS TS") which let us identify an Hifitime Epoch in almost any timescale. 

This blocks any progress towards accurate positioning


rinex/src/epoch/mod.rs
```diff
@@ -122,84 +129,88 @@ pub(crate) fn format(epoch: Epoch, flag: Option<EpochFlag>, t: Type, revision: u
...
-            // two digit issues (old rinex format)
-            if y < 80 {
-                // RINEX did not exist
-                // modern file (2000+) that uses old revision,
-                y += 2000;
-            } else {
-                y += 1900; // [1980:2000]
-            }
+
+    let mut y = i32::from_str_radix(items[0], 10)
```
lnicola, line 143: Not sure how much it matters, but you can repeatedly call `Iterator::next` instead of `collect`, in order to avoid the allocation for `items`.

lnicola (approved):

rinex/src/epoch/mod.rs
gwbres: Oh you mean how we manage `&str.split_ascii_whitespace()` I presume.

We have that in several locations and it could be significant if they get all improved.

It's just I don't a "clean" way to deal for example in this case, when parsing a datetime when you need to collect 6 or 7 fields, I find using `iter.next()` is very verbose. For example : 

```rust
let y_str = iter.next()
    .unwrap();
let y = u32::parse();

let m_str = iter.next()
   .unwrap();
let m = u32::parse();
```

to reduce the code quantity we should have something like `while let Some(str) = iter.next()`  but we need a robust way to identify which field we're about to parse

gwbres: I managed to get it to work :) I
it's not that neat, because I have a custom errors like `Error::YearField` or `Error::MonthField` that makes debugging very easy, that I think we should preserve. Considering how RINEX differs in its format.

I'll see in the next few weeks, if we can improve other spots where I used `split.ascii_whitespace().collect()`

lnicola: Right. But you can still do something like:

```rust
    let mut it = content.split_ascii_whitespace();

    let item = it
        .next()
        .ok_or_else(|| ParsingError::YearField("missing year".to_string()))?;
    let mut y = item
        .parse::<i32>()
        .map_err(|_| ParsingError::YearField(item.to_string()))?;
    /* old RINEX problem: YY is sometimes encoded on two digits */
    if y < 100 {
        if y < 80 {
            y += 2000;
        } else {
            y += 1900;
        }
    }

    let item = it
        .next()
        .ok_or_else(|| ParsingError::MonthField("missing month".to_string()))?;
    let month = item
        .parse::<u8>()
        .map_err(|_| ParsingError::MonthField(item.to_string()))?;
```

Or, with a small helper, a variation like:

```rust
    fn parse_number<'a, T: FromStr<Err = FSE>, FSE, E>(
        s: &'a str,
        invalid: impl FnOnce(&'a str) -> E,
    ) -> Result<T, E> {
        s.parse::<T>().map_err(|_| invalid(s))
    }

    let mut it = content.split_ascii_whitespace();

    let mut y = parse_number(
        it.next()
            .ok_or_else(|| ParsingError::YearField("missing Year".to_string()))?,
        |s| ParsingError::YearField(s.to_string()),
    )?;
    /* old RINEX problem: YY is sometimes encoded on two digits */
    if y < 100 {
        if y < 80 {
            y += 2000;
        } else {
            y += 1900;
        }
    }
    let mut m = parse_number(
        it.next()
            .ok_or_else(|| ParsingError::MonthField("missing month".to_string()))?,
        |s| ParsingError::MonthField(s.to_string()),
    )?;
```

PR #163, opened 2023-09-09
Release 0.13.0

gwbres:
- [x] Rinex 0.13.0
- [x] Sinex 0.2.1 : minor benefits, from v0.13.0 improvements

Improved overall Parsing error reporting: use less generic errors like std::float::error or num::int:error,
use meaningful errors instead: this makes debugging much easier.

New features
  - [x] DCB compensations are now correctly parsed and user API 
(in case this file came with such compensation)
  - [x] PCV compensations are now correctly parsed and introduce their user API
(in case this file came with such compensation)
  - [x] Good progress towards High Precision RINEX support. Although we cannot test it due to unavailable data.
Scaling factors are now supposedly correctly parsed, at least if they fit on one line. Update the (phase data) API to take possible scaling into account (high precision context)

### DIFF
--- a/rinex/src/observation/mod.rs
+++ b/rinex/src/observation/mod.rs
@@ -97,18 +97,6 @@ impl std::fmt::Display for Crinex {
     }
 }
 
-/// DCB Compensation description
-#[derive(Debug, Clone, Default, PartialEq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct DcbCompensation {
-    /// Program used for DCBs evaluation and compensation
-    pub program: String,
-    /// Constellation to which this compensation applies to
-    pub constellation: Constellation,
-    /// URL: source of corrections
-    pub url: String,
-}
-
 /// Describes known marker types
 /// Observation Record specific header fields
 #[derive(Debug, Clone, Default, PartialEq)]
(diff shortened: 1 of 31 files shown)


rinex/src/observation/mod.rs
```diff
@@ -108,69 +108,31 @@ pub struct HeaderFields {
...
-        let mut s = self.clone();
-        if let Some(scalings) = s.scalings.get_mut(&c) {
+    /// Insert a data scaling
+    pub(crate) fn insert_scaling(
+        &mut self,
+        c: Constellation,
+        observable: Observable,
+        scaling: u16,
+    ) {
+        if let Some(scalings) = self.scalings.get_mut(&c) {
```
lnicola, line 123: You can use https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_default to avoid a double lookup here.

lnicola (approved):

larsnaesbye (approved):

rinex/src/observation/mod.rs
gwbres: > You can use https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_default to avoid a double lookup here (but a `HashMap<(Constellation, Observable), u16>` might be better).

The double key (Constellation, Observable) is a good idea, simplifies the overall operation. 
I'll see if I can apply that kind of simplification some place else, where we have multi dimensional arrays

PR #165, opened 2023-09-14
Develop

gwbres:
- [x] Load Directories and files pool recursively.
Now that we have real geodesics operations, it gets tedious to load complex data context with the previous command line interface (one flar per individual files).
This patch enhanced the Cli to load folders recursively.
The only assumption is data being sorted by File Types (one type per folder) see updated docs.

- [x] unify SV position API to kilometer units.
km is much better than meters when dealing with SV.
It is also the direct unit used in Broadcast Ephemeris and SP3 : this solves one conversion

- [x] |Ephemeris - SP3| residual error feature.
We already have the ability to load both SP3 (High Precision/laboratory) Orbits and received Broadcast Ephemeris (from real GNSS signals). One useful analysis is the residual errors between the prediction and the real world.
Currently we can only analyze residual positions errors


larsnaesbye (approved):
Looks good. I encountered some dependency bugs with sp3 before :)

lnicola (approved):

rinex-cli/src/cli.rs
```diff
@@ -448,15 +463,30 @@ Refer to README"))
...
-    }
-    pub fn atx_paths(&self) -> Option<ValuesRef<'_, String>> {
-        self.matches.get_many::<String>("atx")
+    /*
+     * Returns possible name of directory passed
+     * as a specific data provider
+     */
+    pub fn data_dir(&self, key: &str) -> Option<ReadDir> {
+        if let Some(matches) = self.matches.get_many::<String>(key) {
+            if matches.len() == 1 {
```
lnicola, line 472: This here might be a little confusing for the user.

rinex-cli/src/main.rs
```diff
@@ -69,98 +68,177 @@ pub fn create_workspace(path: PathBuf) {
...
+ * Appends directory content, to data context being constructed
+ */
+fn load_rinex_directory(dir: ReadDir, ctx: &mut QcExtraData<Rinex>, ftype: RinexType) {
+    for entry in dir {
+        if let Ok(entry) = entry {
+            let path = entry.path();
+            if path.is_dir() {
+                /* recursive loading */
+                if let Ok(dir) = path.read_dir() {
+                    load_rinex_directory(dir, ctx, ftype);
```
lnicola, line 150: Might as well use [`walkdir`](crates.io/crates/walkdir) here.

gwbres:
> Looks good. I encountered some dependency bugs with sp3 before :)

"workspace" is Cargo does not help :)

PR #168, opened 2023-09-15
Integrate SP3 lib to ecosystem

gwbres:
One solution to having a single cargo workspace for this ecosystem is either

* integrate SP3 into this crate: which facilitates future developments (use cargo local referencing)
* move all applications (non libraries, ie., library users like rinex-cli) into an external workspace

Having SP3 into this is probably the best solution.   
For example `rinex-qc` makes no sense without SP3 and RINEX knowledge.   
`rinex-cli` is a heavy user of `rinex-qc` (one of the two major opmodes). 
Also, new developments (advanced geodesics), calls for `rinex-cli` using local referencing.

I think that answers the question.


lnicola:
Note that you can use `git-subtree` to move a git repository to another one, while keeping the history intact.

.gitignore
```diff
@@ -1,10 +1,16 @@
+debug/
```
lnicola, line 1: Is this needed?

lnicola (approved):

lnicola:
> while keeping the history intact

image

Okay, maybe let's not bother :joy:.

.gitignore
gwbres: that repo was 100% initiated with the proposed .gitignore generated by github, and I copied this from that.
I don't think it's actually needed because /target/debug is the actual location

gwbres:
> Note that you can use [`git-subtree`](https://doc.rust-lang.org/nightly/clippy/development/infrastructure/sync.html) or [`josh`](https://github.com/josh-project/josh) to move a git repository to another one, while keeping the history intact.
> 
> I've used the former, but can't say it's very pleasant.

good to know 👍  Probably not needed on this one, considering this repo is almost brand new

.gitignore
lnicola: Not needed. The only thing `cargo new` puts in `.gitignore` is `/target`. I think it also used to add `**/*.rs.bak` and `Cargo.lock` for libraries, but no longer does it.

PR #179, opened 2023-10-13
sinex: Fix deprecation warning of `NaiveDateTime::and_hms` by replacing with `and_hms_opt`.

larsnaesbye:
The [newer implementation](https://docs.rs/chrono/0.4.31/chrono/naive/struct.NaiveDate.html#method.and_hms_opt) doesn't panic but might return `None` instead.
While this patch simply unwraps the result fixing the deprecation warning, it should probably handle `None` separately.

### DIFF
--- a/sinex/src/datetime.rs
+++ b/sinex/src/datetime.rs
@@ -17,7 +17,7 @@ pub fn parse_datetime(content: &str) -> Result<chrono::NaiveDateTime, ParseDateT
     let h = secs / 3600.0;
     let m = (secs - h * 3600.0) / 60.0;
     let s = secs - h * 3600.0 - m * 60.0;
-    Ok(dt.and_hms(h as u32, m as u32, s as u32))
+    Ok(dt.and_hms_opt(h as u32, m as u32, s as u32).unwrap())
 }
 
 #[cfg(test)]


sinex/src/datetime.rs
```diff
@@ -17,7 +17,7 @@ pub fn parse_datetime(content: &str) -> Result<chrono::NaiveDateTime, ParseDateT
     let h = secs / 3600.0;
     let m = (secs - h * 3600.0) / 60.0;
     let s = secs - h * 3600.0 - m * 60.0;
-    Ok(dt.and_hms(h as u32, m as u32, s as u32))
+    Ok(dt.and_hms_opt(h as u32, m as u32, s as u32).unwrap())
```
gwbres, line 20:
```rust
Ok(dt.and_hms_opt(h as u32, m as u32, s as u32)?)
```

in relation to `ParseDateTimeError`

larsnaesbye: Done now :)

larsnaesbye: But this fails with `use .ok_or(...)? to provide an error compatible with `Result<NaiveDateTime, ParseDateTimeError>`

gwbres: you can either use thiserror::Error to provide automated casts from chrono::Errors to ParseDateTimeError, 
or else need to cast yourself, with .ok_or(Err(Error::SomeKindOfError)).

you can find similar scenarios in the rinex/src/epoch or rinex/src/header modules

I would not recommend spending time on this, this crate will eventually get updated to use Hifitime

gwbres:
Right now I see the SINEX topic is at the very bottom of the planning.
See the explanations down below, let me know what you think and if you have any input 

I'm currently working on precise positioning, which basically uses Obs/Nav/SP3 and soon ATX.
Therfore, amongst exotic format, ATX is the next one that we will finalize and integrate in a workflow. 

So that means I can only work on SINEX once this point is reached.
More importanly, it is still not clear to me how SINEX may be integrated to the processing workflow (positioning mostly)

PR #184, opened 2023-11-24
v0.15.1

gwbres:
- [x] starting a wiki page with tutorials
- [x] improved the ci script
- [x] rnx2crx V2.2.0 significant improvements
- [x] rinex v0.15.1 minor changes
  - [x] Observation Data generation has been fixed 
- [x] rinex-cli v0.9.6 improved positioning opmode
- [x] rnx2cggtts v1.0 new binary :) 
- [x] trying to fix the release scripts


gwbres:
Ok so I can provide a new autobuilder for all apps,  
the CI script gets also improved, we test the build of all binaries on Windows and MAC on each PR.
I will push this today or tomorrow.

Happy to soon forget about this topic

larsnaesbye:
Will there be prebuilt binaries for both major macOS architectures? If not, then I suggest prioritizing aarch64 as x86_64 is being phased out.

gwbres:
Hello @larsnaesbye,

> Will there be prebuilt binaries for both major macOS architectures? If not, then I suggest prioritizing aarch64 as x86_64 is being phased out.

you can take a look at my _cggtts_ library with its cli application.  
The upload does not seem to properly work at the moment,
I focus on ubuntu-latest, windows-latest and mac-os latest as the destination targets (standard PC OS).
I'm not sure what the second macos architecture would be then ?

larsnaesbye:
Apple is at the end of a hardware transition from Intel to ARM architecture. The default Github runners are still using the now older Intel architecture. While binaries will run on newer Macs, they will run in emulation mode and therefore slower.

I think it's a easy as setting `runs-on: macosx-latest-xlarge` like they indicate [here ](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/).

larsnaesbye:
On second thought, let's keep to `macos-latest`. It gives maximum compatibility, and when the runners get updated to ARM then people will get faster binaries. If we want to, we can just make a note about it :)

gwbres:
https://github.com/gwbres/cggtts/releases 

that I will port to the other projects like this one :)

gwbres:
@larsnaesbye 

the binaries work fine. But this created a behavior I did not see coming, I usually use the Cargo.toml location to determine a default workspace, but that no longer stands in the case of a binary, or simply "cargo install" by itself.  
This makes the `-w --workspace` flag mandatory in this case

PR #186, opened 2023-11-28
working on release script

gwbres:


larsnaesbye:
I still think it would help to change line 46 to target `aarch64-apple-darwin` 

https://github.com/georust/rinex/blob/main/.github/workflows/release.yml#L46

lnicola (approved):

larsnaesbye (approved):

gwbres:
> I still think it would help to change line 46 to target `aarch64-apple-darwin`
> 
> https://github.com/georust/rinex/blob/main/.github/workflows/release.yml#L46

you mean add this one too, because they're two different things.  
`aarch64-apple-darwin`  is a embedded / cross-compiled target and `86_64-apple-darwin` is standard OS, right ?

gwbres:
> Those targets look a bit messy (musl vs. GNU, those binutils `strip`s vs. could be replaced with `CARGO_PROFILE_RELEASE_STRIP`, is anyone using 32-bit binaries? etc.).

good to know, I'll try to remove the manual `strip`.
What is wrong with musl + GNU though ?

lnicola:
> so if I understand correctly, we can remove the manual strip and have cargo do it instead, for example by putting the following inside the Cargo.toml workspace

Yes, but every profile option can also be set using environment variables, like in https://doc.rust-lang.org/cargo/reference/environment-variables.html. That would probably be a better fit here.

> What is wrong with musl + GNU though ?

Nothing, I just wouldn't guess that 64-bit builds use musl and 32-bit ones use glibc.

larsnaesbye:
> > I still think it would help to change line 46 to target `aarch64-apple-darwin`
> > https://github.com/georust/rinex/blob/main/.github/workflows/release.yml#L46
> 
> you mean add this one too, because they're two different things. `aarch64-apple-darwin` is a embedded / cross-compiled target and `86_64-apple-darwin` is standard OS, right ?

At the moment macOS runs on the older (no longer manufactured) x86_64-apple-darwin architecture as well as the newer machines using aarch64-apple-darwin architecture. On all newer machines software can run natively using the aarch64-apple-darwin architecture, or in [Rosetta 2 emulation](https://en.wikipedia.org/wiki/Rosetta_(software)#Rosetta_2) using the x86_64-apple-darwin architecture. So the 'standard' OS is aarch64-apple-darwin.

gwbres:
@lnicola 

I confirm that we reduce the binaries to the same size; now with cargo instead of a manual `stripping`

gwbres:
> > so if I understand correctly, we can remove the manual strip and have cargo do it instead, for example by putting the following inside the Cargo.toml workspace
> 
> Yes, but every profile option can also be set using environment variables, like in https://doc.rust-lang.org/cargo/reference/environment-variables.html. That would probably be a better fit here.
> 
> > What is wrong with musl + GNU though ?
> 
> Nothing, I just wouldn't guess that 64-bit builds use musl and 32-bit ones use glibc.

OK, sorry i'm not super familiar with all of this.  
I think we should only target 64b desktop environment with the auto shipment.   
It makes much more sense to manually cross-compile on exotic architectures.

lnicola:
At this point, 32-bit x86 is pretty exotic. Anyway, musl vs. glibc isn't necessarily about cross-compilation, but rather:

 - glibc is dynamically linked but versioned, so if you compile a binary with glibc 2.38, it will run on 2.38 and future versions
 - musl is statically linked, so a musl binary will run on every system with the same architecture

However, musl is less advanced in some ways. The memory allocator, for example, used to be very slow. It's better these days, but I hear that e.g. musl GDAL is still much slower than the glibc one.

I don't know if any of this matters for `rinex`, but the current setup is somewhat surprising.

gwbres:
The build works on linux and both windows, it's just failing to locate the products.
There is a link issue on mac that I will not investigate.
This script should probably be allowed to fail and proceed

gwbres:
> Nothing, I just wouldn't guess that 64-bit builds use musl and 32-bit ones use glibc.

is x86_64-unknown-linux-gnu the standard target then ? looks like what I have on my x86_64 PC

PR #188, opened 2023-11-30
Develop

gwbres:
- bump clap to 4_4_10
- introduce `lsqspp` new positioning method 
- plot PVT::velocity

### DIFF
--- a/rinex-cli/src/cli.rs
+++ b/rinex-cli/src/cli.rs
@@ -1,10 +1,13 @@
 use clap::{Arg, ArgAction, ArgMatches, ColorChoice, Command};
-use gnss_rtk::prelude::Config;
 use log::{error, info};
-use rinex::prelude::*;
-use rinex_qc::QcOpts;
+
 use std::path::Path;
 use std::str::FromStr;
+
+use rinex::prelude::*;
+use rinex_qc::QcOpts;
+
+use gnss_rtk::prelude::{Config, Mode as SolverMode};
 
 pub struct Cli {
     /// Arguments passed by user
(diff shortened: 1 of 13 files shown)


rinex-cli/src/cli.rs
```diff
@@ -440,12 +453,22 @@ Primary RINEX was either loaded with `-f`, or is Observation RINEX loaded with `
     pub fn quiet(&self) -> bool {
         self.matches.get_flag("quiet")
     }
-    /// Returns true if SPP position solver is enabled
-    pub fn spp(&self) -> bool {
-        self.matches.get_flag("spp")
+    /* returns RTK solver mode to implement */
+    pub fn solver_mode(&self) -> Option<SolverMode> {
```
lnicola, line 457: I'd make this return `SolverMode`, like the other `solver_mode` does.

lnicola (approved):

rinex-cli/src/cli.rs
gwbres: That is correct 👍  thanks 
Now that we have several positioning opmodes, we can define a default one.
The default one is now `lsqspp` because it gives the best results, and I think these tools should always prefer the most performant (resolution wise, not computation wise) options

gwbres: hum.. I understand why it was proposed this way 😆 sorry for the misunderstanding.

Post process Positioning in `rnx2cggtts` is always required, hence a strategy always has to be defined.  

In `rinex-cli` (at least current approach), it is one opmode amongst others. 
In other words, the position solver may not deploy (especially, considering it is a lengthy and computation heavy process). 

Therefore we need someway to describe all of that, currently we have :
 
 - `--spp` to deploy the +/- 5m strategy
 -  `--lsqspp `  would deploy the +/- 1m strategy
 - `--ppp` will later come to deploy he +/- 1 cm strategy
 -  combine `-p ` do disable record analysis (focused on positioning)
 - the Option::None then tells us the solver is not to be deployed

PR #190, opened 2023-12-01
Ci release

gwbres:


.github/workflows/release.yml
```diff
@@ -79,21 +75,21 @@ jobs:
...
-          ublox-rnx="target/${{ matrix.target }}/release/ublox-rnx.exe"
+          ubloxrnx="target/${{ matrix.target }}/release/ublox-rnx.exe"
         else
           crx2rnx="target/${{ matrix.target }}/release/crx2rnx"
           rnx2crx="target/${{ matrix.target }}/release/rnx2crx"
-          rinex-cli="target/${{ matrix.target }}/release/rinex-cli"
+          rinexcli="target/${{ matrix.target }}/release/rinex-cli"
           rnx2cggtts="target/${{ matrix.target }}/release/rnx2cggtts"
-          ublox-rnx="target/${{ matrix.target }}/release/ublox-rnx"
+          ubloxrnx="target/${{ matrix.target }}/release/ublox-rnx"
```
lnicola, line 86: Doesn't the target directly get doubled? It's both here and in the cp below.

gwbres: Hello,
not sure I understand what you mean.
We have the build step, and then I copy the generated binaries into an "archive" folder, to be zipped later on

lnicola: `$crx2rnx` includes `target/whatever/release`, it gets stored into `$CRX2RNX`, then you do `cp target/whatever/release/$CRX2RNX`.

gwbres: 😆 so obvious I did not see it. Thanks ! 👍 
In this current state, the binaries will exist and will be zipped, the release gets created,
but we still need to retrieve the zipped binaries and use a job dedicated to release upload.

The way [I did it at the bottom of this script](https://github.com/gwbres/cggtts/blob/main/.github/workflows/release.yml), is to download all zipped artifacts and then trigger the upload job

lnicola (approved):

lnicola:
I wonder if a single app-approach (`rinex crx2rnx [stuff]`) wouldn't be easier for the users.

If you need more inspiration for CI workflows, you can check out https://github.com/rust-lang/rust-analyzer/blob/master/.github/workflows/release.yaml and https://github.com/rust-lang/rust-analyzer/blob/master/.github/workflows/ci.yaml. They're not perfect, but maybe you see something interesting in there.

gwbres:
> I wonder if a single app-approach (`rinex crx2rnx [stuff]`) wouldn't be easier for the users.
> 
> If you need more inspiration for CI workflows, you can check out https://github.com/rust-lang/rust-analyzer/blob/master/.github/workflows/release.yaml and https://github.com/rust-lang/rust-analyzer/blob/master/.github/workflows/ci.yaml. They're not perfect, but maybe you see something interesting in there.

Thank you for pointing that out, I will take a look if we can grab a few things from there.  

Regarding the app, honestly I don't know. It's only a matter of writing a different name in a workflow script.

gwbres:
> If you need more inspiration for CI workflows, you can check out https://github.com/rust-lang/rust-analyzer/blob/master/.github/workflows/release.yaml and https://github.com/rust-lang/rust-analyzer/blob/master/.github/workflows/ci.yaml. They're not perfect, but maybe you see something interesting in there.


From what I can see we're not far from truth, they also use a very lengthy and tedious artifact upload then download then publish sucessive steps. Unfortunately the publish step is what it is, and the artifact download seems the only option to retrieve what you're interested in.  I may try to use a `strategy.matrix` in the future, to reduce the length of the script.

I'll trigger this release to make sure the binaries name are correct.   
Then we need one final PR for the artifacts actual publication

Issue #103, opened 2023-07-24
ANTEX Frequency parsing does not support all available values


jbangelo:
The ANTEX record parsing does not properly handle all of the frequency bands available in current ANTEX files from IGS. I found [this](https://files.igs.org/pub/data/format/antex14.txt) description the ANTEX format, and in the description of the `START OF FREQUENCY` section it lists several potential frequency bands that could be encoded but the current parsing doesn't handle all of these cases. The current implementation appears to label anything it doesn't expect as either `Carrier::L1` or `Carrier::E1`. It also appears that all BeiDou frequency bands are given Galileo `Carrier` values.

This behavior means it's impossible to differentiate which is the "real" L1 calibration value and which are the mislabeled values. Here is a minimal program, the output shows that the list of frequencies for at least some antennas are ambiguous since there are multiple entries with the same carrier value. I'm also attaching the ANTEX file that I'm using locally, but any relatively recent [file from IGS](https://igs.org/wg/antenna/) demonstrates this issue.

```rust
fn main() {
    let file_path = "/path/to/igs14.atx";

    let rnx = rinex::Rinex::from_file(&file_path).unwrap();
    if let rinex::record::Record::AntexRecord(atx) = rnx.record {
        for (antenna, frequencies) in atx.iter() {
            println!("Antenna {} {}", antenna.ant_type, antenna.sn);
            for freq in frequencies.iter() {
                println!("\tFrequency {:?}", freq.carrier);
            }
        }
    }
}
```

[igs14.atx.zip](https://github.com/georust/rinex/files/12152426/igs14.atx.zip)

gwbres:
@jbangelo,

Thank you so much for the great feedback !

>in the description of the START OF FREQUENCY section it lists several potential frequency bands that could be encoded but the current parsing doesn't handle all of these cases

I'll try to provide an answer to this in August

gwbres:
Hello @jbangelo,

So the classifcation is correct.

Basically you're saying that the "carrier" signal information, which we can relate to a frequency band, that is attached to each individual `Frequency` in a sorted ATX record, is regularly incorrect (bad parsing).

It solely relies on [this method](https://github.com/georust/rinex/blob/main/rinex/src/carrier.rs#L548C32-L548C32), which  was developed for that ATX requirement specifically. The carrier struct has convenient frequency band methods (mostly for calculations).

It just looks like it's missing one or two cases to identify (E5b mostly, LEX is L6 in our lib).
[...]

PR #194, opened 2023-12-26
Develop

gwbres:
- [x] #103 
- [x] `antex` feature
- [x] improved plots x axis: use MJD 
newplot(10)
- [x] RINEX3 ionod corrections and visualization method 
- [x] Corrected all GNSS combinations
newplot(15)
newplot(18)


rinex/src/observation/record.rs
```diff
@@ -1461,13 +1441,13 @@ fn dual_freq_combination(
                 };
 
                 let value = match combination {
-                    Combination::NarrowLane => alpha * beta * (v_i + gamma * v_j),
-                    _ => alpha * beta * (v_i - gamma * v_j),
+                    Combination::NarrowLane => alpha * (beta * v_i + gamma * v_j),
```
lnicola, line 1444: This is intended, right?

rinex/src/antex/record.rs
```diff
@@ -203,25 +412,25 @@ impl Merge for Record {
...
-                                ffrequencies.push(frequency.clone());
-                            }
-                        }
-                    }
-                }
-            } else {
-                self.push(antenna.clone());
-            }
-        }
+    fn merge_mut(&mut self, _rhs: &Self) -> Result<(), merge::Error> {
```
lnicola, line 415: Is this supposed to do nothing?

lnicola:
`rinex/src/antex/antenna/cospar.rs` is an empty file, probably not intended.

rinex/src/observation/record.rs
gwbres: Correct, the implemented equation is simplified

gwbres:
> `rinex/src/antex/antenna/cospar.rs` is an empty file, probably not intended.

Correct, it is declared in src/antex/antenna/sv.rs, SV ou RX being the two types of antenna we can have.
Thank you for catching this

lnicola (approved):

rinex/src/antex/record.rs
gwbres: hum, good catch once again. I forgot to re-enable this.
Reworking ATX requires to rework this method, I just forgot about it.
I will actually push one or two tests for this too.

FYI for each RINEX format, the idea is to support the Merge trait, to merge RINEX B into RINEX A, if their types matches. Basically, this is a fileops that was historically provided by teqc. 
It turns out to actually be mandatory in RINEX post processing and I use it all the time.
RINEX post processing requires putting toghethers entire datasets. One RINEX spans 24h of data. With Merge we can put a dataset together, or even cover 48h of data for example. A RINEX by itself is not enough for navigation, you need at least 3 or 4 of them.

PR #198, opened 2023-12-28
Cli improvements

gwbres:
Convert existing (RINEX-cli) opmodes into `clap::Subcommands`.

The application has several opmodes that are exclusive.   
Subcommands which are exclusive by nature, are the ideal way to depict them. 

Subcommands then has several benefits

- it makes it easier to understand the CLI with one common panel and then subcommands
- each panel (including the subcommands) have their own "help" menu. So it makes it easier to document and easier to understand
- this drastically reduce the quantity of information we have to display:

I merged the `RNX2CGGTTS` binary into a RINEX-cli -p -cggtts opmode. 
Indeed, this one uses the -p behavior, with a customized output. The previous implementation had the code duplicated.
It also required to update both applications on any dependency upgrade.

I renamed the previous "default opmode" to "--graph -G". Considering it's more data visualization oriented.

I'm also introducing a script to test the major applications. Since we're shipping those applications online now, it is almost mandatory to have more CLI/applications continuous integration.

##  Details

We only have a couple of commands that are shared accross opmodes:
* the general file loading infrastructure (-f, -d)
* the workspace and env definition (verbosity -q + workspace location -w)
* the RX antenna definition (RX position definition --rx-ecef/--rx-geo)
* the preprocessor: -P to filter data prior running an opmode

The current opmodes are (use -h for each of them):
* -i: identification opmode, documentation initiated in the Wiki.
* --split -s: documentation initiated in the Wiki
* --merge -m: documentation initiated in the Wiki
* --tbin: time binning opmode, documentation initiated in the Wiki
* -Q: quality check. See work in progress in the wiki
* -g: graph mode - See work in progress in the wiki
It's basically the previous ""default"" opmode, that would generate
Record visualization.

image

(1) is the general front panel: basically everything that is shared with other opmodes  
(2) are the opmodes: you have to select one (and can only select one) to actually do something  

Example of subcommand help/documentation : 

image

image

(3): RNX2CGGTTS is now a precise positioning opmode. It makes maintaining this operation easier.
It reduces the code quantity: can share the solver logic. It allows sharing the Solver configuration file -c...


rinex-cli/src/cli/graph.rs
```diff
@@ -0,0 +1,142 @@
...
+                .long("clk")
+                .action(ArgAction::SetTrue)
+                .help("SV clock bias (offset, drift, drift changes)."),
+        )
+        .next_help_heading("Atmosphere conditions")
+        .arg(
+            Arg::new("tropo")
+                .long("tropo")
+                .action(ArgAction::SetTrue)
+                .help("Plot trospheric delay from meteo sensors estimation. Requires METEO RINEX."),
```
lnicola, line 128:
```suggestion
                .help("Plot tropospheric delay from meteo sensors estimation. Requires METEO RINEX."),
```

rinex-cli/src/cli/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+    /*
+     * Creates File/Data context defined by user.
+     * Regroups all provided files/folders,
+     */
+    pub fn from_cli(cli: &Cli) -> Result<Self, Error> {
+        let mut data = RnxContext::default();
+        /* load all directories recursively, one by one */
+        for dir in cli.input_directories() {
+            let walkdir = WalkDir::new(dir).max_depth(5);
```
lnicola, line 83: Should probably be configurable in the long run.

rinex-cli/src/cli/qc.rs
```diff
@@ -0,0 +1,29 @@
...
+See online documentations [https://docs.rs/gnss-rtk/latest/gnss_rtk/prelude/enum.Method.html#variants]."))
+        .arg(
+            Arg::new("cfg")
+                .short('c')
+                .long("cfg")
+                .required(false)
+                .value_name("FILE")
+                .action(ArgAction::Append)
+                .help("Pass a QC configuration file (JSON).
+[] is the structure to represent in JSON.
```
lnicola, line 27: Never seen these before, are they placeholders for something?

lnicola:
`rinex-cli/src/file_generation.rs` looks like another empty file.

rinex-cli/src/graph/record/meteo.rs
```diff
@@ -57,4 +63,41 @@ pub fn plot_meteo(rnx: &Rinex, plot_context: &mut PlotContext) {
...
+
+    let has_wind_direction = theta.len() > 0;
+
+    let rho = rnx
+        .wind_direction()
+        .map(|(t, _)| {
+            if let Some(speed) = rnx
+                .wind_speed()
+                .filter(|(ts, _)| *ts == t)
+                .reduce(|k, _| k)
```
lnicola, line 85: Is this `.next)`? Could probably merged with the `filter` above (using `find`).

rinex-cli/src/positioning/cggtts/post_process.rs
```diff
@@ -0,0 +1,83 @@
...
+            "rinex-cli v{} - https://georust.org",
+            env!("CARGO_PKG_VERSION")
+        ));
+
+    for track in tracks {
+        cggtts.tracks.push(track);
+    }
+
+    let filename = ctx.workspace.join(cggtts.filename());
+    let mut fd = File::create(&filename)?;
```
lnicola, line 78: Nit: didn't check, but this might need a `BufWriter`.

rinex-cli/src/main.rs
```diff
@@ -4,134 +4,53 @@
...
 mod identification; // high level identification/macros
-mod plot; // plotting operations
+mod positioning;
+mod qc; // QC report generator // plotting operations // file operation helpers // graphical analysis // positioning + CGGTTS opmode
 
 mod preprocessing;
 use preprocessing::preprocess;
 
-mod positioning;
+// mod positioning;
```
lnicola, line 16: Can be removed, it's already declared above.

lnicola (approved):
Rubber-stamping this, I can't really review it properly.

rinex-cli/src/cli/mod.rs
gwbres: It does not cost much to add this into the general panel.
Something like `--depth X` to make the `-d` flag more flexible.
I will add it

rinex-cli/src/cli/qc.rs
gwbres: No it's just I'm working on the documentation online and some references do not exist yet

gwbres:
> `rinex-cli/src/file_generation.rs` looks like another empty file.

Hum, I don't have this one on my side.  
It did exist at some point indeed, but it was either merged someplace else and partially replaced by the current `fops.rs`

lnicola:
Right, sorry, the GitHub display is a bit misleading:

image

vs.

image

gwbres:
Yeah indeed, I spotted that one and removed it.
The new fops infrastructure regroups this kind of thing.
It could benefit some more logic for the process of file creation, because we happen to do the same thing in a couple of places.
Will try to improve that aswell

rinex-cli/src/positioning/cggtts/post_process.rs
gwbres: Hello @lnicola,

i'm not sure I understand what you mean here.

Buffered Writing is built in, in our objects. For example when creating a RINEX: it is hidden inside.
Here, we're just creating a file handle, prior writing a file.

CGGTTS file operations are not buffered because these files are very tiny compared to RINEX files.
But that will be improved in the future

lnicola: > CGGTTS file operations are not buffered because these files are very tiny

I didn't check, but I've seen too many programs doing 1-byte IO ops. Anyway, at least you're aware of it.

PR #209, opened 2024-03-06
Context improvements

gwbres:
This PR improves and simplifies the `RnxContext` structure.
It makes it more efficient to represent what is required in GNSS and RINEX post processing.
For example, it allows now to preprocess all types of products we support and generate new products.
It slightly reworks the way we operate the RnxContext structure and removes a big memcopy that was previously required in previous version, making the initial phase of the program more efficient.


gwbres:
Hello @lnicola 

I'm trying to improve the `RnxContext` structure substancially but I need your help to figure out one method specifically.

To give more context:  `RnxContext` is not correctly defined today, in the sense it does not hold internal references to inner RINEX or SP3 files. The consequence of this, is [we need to  `Clone()` the RINEX/SP3 objects we have just parsed](https://github.com/georust/rinex/blob/0b4ea045f5d25b414a765f3b9d56250e3c7a2a92/rinex/src/context/mod.rs#L411), when forming `RnxContext`, which is quite a heavy operation and you ideally want to avoid.

With this PR i'm trying to fix that, mostly by introducing a lifetime to `RnxContext`. I am very close to a working solution

rinex/src/context/mod.rs
```diff
@@ -24,561 +22,268 @@ use rinex_qc_traits::HtmlReport;
...
-                match ret.load(&path) {
-                    Ok(_) => trace!("loaded \"{}\"", filename),
-                    Err(e) => error!("failed to load \"{}\", {:?}", path.display(), e),
-                }
-            }
+impl<'a> BlobData<'a> {
+    /// Returns reference to inner RINEX data.
+    pub fn as_rinex(&self) -> Option<&'a Rinex> {
+        match self {
+            Self::Rinex(r) => Some(r),
```
gwbres, line 82: @lnicola 

The hidden `BlobData` struct contains internal references to either RINEX or SP3.
Retrieving one of these references works perfectly

rinex/src/context/mod.rs
```diff
@@ -24,561 +22,268 @@ use rinex_qc_traits::HtmlReport;
...
-        } else if let Some(data) = &self.atx {
-            Some(data)
-        } else if let Some(data) = &self.clk {
-            Some(data)
-        } else {
-            None
+    /// Returns mutable reference to inner RINEX data.
+    pub fn as_mut_rinex(&mut self) -> Option<&'a mut Rinex> {
+        match self {
+            Self::Rinex(r) => Some(r),
```
gwbres, line 89: @lnicola 

But in processing workflow, you sometimes need mutable access, for example to rework the context prior some algorithm. This method is trying to provide such access, but it fails to compile.
That's because `r` here has an actual `&& Rinex` type I think.

How do you solve this ? I think we should use some kind of fancy structure, like `RefCell<Rinex>` in BlobData and not simply `&Rinex`

rinex/src/context/mod.rs
```diff
@@ -24,561 +22,268 @@ use rinex_qc_traits::HtmlReport;
...
-    pub meteo: Option<ProvidedData<Rinex>>,
-    /// Optional IONEX for accurate Ionosphere compensation.
-    pub ionex: Option<ProvidedData<Rinex>>,
+#[derive(Clone)]
+enum BlobData<'a> {
+    /// RINEX content
+    Rinex(&'a Rinex),
+    /// SP3 content
+    Sp3(&'a SP3),
 }
```
gwbres, line 76: @lnicola 

This is the new structure trying to remove the need to Clone internal structures

lnicola:
Sorry, I didn't have a chance to look at this yesterday, and now I'm pretty confused. Even when looking 91664b2569283092dcd60d6755394b37e677ea4b and only the `rinex` crate, I get quite a lot of build errors:

image

lnicola:
But yeah, looking at that pattern, it might need a `Rc<RefCell<..>>`, which is a bit awkward to work with. Are you sure that `BlobData` shouldn't "own" `Rinex` and `Sp3`? Can a `Rinex` value be a part of more than one `BlobData`?

gwbres:
> But yeah, looking at that pattern, it might need a `Rc<RefCell<..>>`, which is a bit awkward to work with. Are you sure that `BlobData` shouldn't "own" `Rinex` and `Sp3`? Can a `Rinex` value be a part of more than one `BlobData`?

Thank you once again for all your input, it turns out it is the "natural" conclusion I came with this morning, that I have injected with the latest commits.

The conclusion is we have a solution to avoid this heavy memcopy, but it requires to slightly modify the way we operate/build/define the context at high level (in the applicatation itself)

rinex/src/context/rtk.rs
gwbres, line 1: RTK and differential analysis are not supported yet. We will do that later, most likely when the position solver supports RTK/differential solving method itself

rinex/src/context/mod.rs
gwbres, line 1: moved into a single file

larsnaesbye (approved):

PR #212, opened 2024-03-18
OBS_RINEX formatting: fix tiny issues when formatting header section

gwbres:


rinex/src/header.rs
```diff
@@ -1613,8 +1613,11 @@ impl Header {
                     for (constell, observables) in &obs.codes {
                         let mut descriptor = String::new();
                         descriptor.push_str(&format!("{:x}{:5}", constell, observables.len()));
-                        for observable in observables {
-                            descriptor.push_str(&format!(" {}", observable));
+                        for (i, observable) in observables.iter().enumerate() {
+                            if (i % 13) == 0 && (i > 0) {
+                                descriptor.push_str(&format!("        ")); // TAB
```
lnicola, line 1618:
```suggestion
                                descriptor.push('\t');
```

lnicola (approved):

rinex/src/header.rs
gwbres: hum, I don't think this would work considering the tab length is not something standard right ?
And I doubt the choice they made in the RINEX format will follow what a "standardized" tab is.
But I'll give it a try, we have a bunch of those in the library

lnicola: Ah, okay, then `.push_str("        ")` should work. I noticed you had spaces in there and assumed it used to be a tab, but your editor replaced it.

larsnaesbye (approved):

rinex/src/header.rs
gwbres: I will give `\t` a try in my future PR.
Like I said, this will probably work in a couple of places, but unfortunately RINEX is not consistent and the tabulation width varies a lot

PR #215, opened 2024-03-23
Fix rinex-cli interpretation of rx-geo

jigpu:
The rx-geo parameter expects lat/lon in degrees, but the geodetic2ecef function wants them in radians. This leads to positions being wildly incorrect. This commit switches to calling the GroundPosition::from_geodetic function instead, which is happy to directly recieve the lat/lon/alt tuple as-is without conversion.

### DIFF
--- a/rinex-cli/src/cli/mod.rs
+++ b/rinex-cli/src/cli/mod.rs
@@ -11,8 +11,6 @@ use rinex::prelude::*;
 
 use crate::fops::open_with_web_browser;
 
-use map_3d::{geodetic2ecef, Ellipsoid};
-
 // identification mode
 mod identify;
 // graph mode
@@ -296,8 +294,8 @@ Otherwise it gets automatically picked up."))
     pub fn manual_position(&self) -> Option<(f64, f64, f64)> {
         if let Some(position) = self.manual_ecef() {
             Some(position)
-        } else if let Some((lat, lon, alt)) = self.manual_geodetic() {
-            Some(geodetic2ecef(lat, lon, alt, Ellipsoid::WGS84))
+        } else if let Some(position) = self.manual_geodetic() {
+            Some(GroundPosition::from_geodetic(position).to_ecef_wgs84())
         } else {
             None
         }


gwbres:
Hello @jigpu 

Good catch, thanks for your contribution once again.
I only used reference positions that are predefined in the file header so far, thats why this had not been caught yet

gwbres (approved):

jigpu:
@gwbres No need to change the schedule on my account. I'm literally just blindly playing around in waters far beyond my depth and reporting issues along the way :sweat_smile: I learned about "RINEX" only a week ago after looking into Android's raw GPS ability. I've been doing a lot of bumbling around in a (perhaps overly-ambitious :laughing:) attempt to get something to turn the data recorded from my phone into a position.

I'm still having trouble getting the rinex-cli position solver to do much with that data. It'll happily* correctly solve when I give it 24o, 24n, and sp3 files from a local reference station, but not so much if I substitute out the 24o with one from my phone. I suspect the phone's data might be too poor for rinex-cli to handle, though I could also just be doing something wrong. (I was able to get RTKLIB to spit out a solution, but I'm under no illusions that the phone's data is going to be pretty terrible compared to anything you'd normally feed to rinex-cli).


*Actually, it has a *small* issues with the data from the reference station. It spits out a `invalid Gregorian date: Carry` error unless I truncate the file slightly. There's a "RINEX FILE SPLICE" near the end that I think rinex-cli doesn't like...

gwbres:
>  I've been doing a lot of bumbling around in a (perhaps overly-ambitious 😆) 

I can only encourage ! If you're motivated enough, let's just switch to our [Discord channel (you can join in for free and tie it to your github account for instance)](https://discord.com/invite/Fp2aape), it will be better suited to discuss complex topics

> I'm still having trouble getting the rinex-cli position solver to do much with that data [..]
>  but not so much if I substitute out the 24o with one from my phone

Position solving is the most advanced topic, I'm the app author and probably the only person that has used this opmode to this day, so this is not surprizing. If I get it right, you manage to get your phone spit out its signal observations as a RINEX (Observation) file. 

> It'll happily* correctly solve when I give it 24o, 24n, and sp3 files from a local reference station

24o are the signal observations, 24n are the ephemerides and SP3 are high precision orbits.   
Keep in mind you need to gather those for that very same day, otherwise it will not work.
You can register for free on the NASA/CDDIS portal, to retrieve some of these files. 
Ideally Rinex-cli should be able to resolve with a simpler context (24o+24n), but we're unable to do that. 
I will try to fix that soon

> There's a "RINEX FILE SPLICE" near the end that I think rinex-cli doesn't like...

Do you care enough to share an example of your Observation Data so I can reproduce on my side ?  
I think `RINEX FILE SPLICE` is the special (this is not standard) marker  that the `teqc` program places inside a RINEX that was either merged or truncated, at the instant it did that job

> Actually, looks like even the data from the reference station just shows a single point at the apriori position... I must definitely be doing something wrong

My recommendation would be to take advantage of the logger ( `$RUST_LOG`), the application is pretty verbose and will point out what step is failling in the solving process. You can directly copy and paste its output here, or on Discord, I will rapidly understand what issues we're facing

PR #220, opened 2024-03-24
OBS V3 formatting

gwbres:
- [x] fixes minor issue that caused OBS(V3) formatting to be wrong
- [x] unlock dedicated test
- [x] correct the timestamp that we format as `TIME_OF_FIRST OBS` and `TIME_OF_LAST OBS`

### DIFF
--- a/rinex/src/observation/record.rs
+++ b/rinex/src/observation/record.rs
@@ -623,7 +623,7 @@ fn fmt_epoch_v3(
         }
         lines.push('\n');
     }
-    lines
+    lines.trim_end().to_string()
 }
 
 fn fmt_epoch_v2(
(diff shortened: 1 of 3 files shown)


rinex/src/observation/record.rs
```diff
@@ -623,7 +623,7 @@ fn fmt_epoch_v3(
         }
         lines.push('\n');
     }
-    lines
+    lines.trim_end().to_string()
```
lnicola, line 626:
```suggestion
    lines.truncate(lines.trim_end().len());
    lines
```

(`lines.pop()` is also useful at times)

lnicola (approved):

PR #221, opened 2024-03-25
switch from chrono::Duration to chrono::TimeDelta

larsnaesbye:
Duration is now simply an alias of TimeDelta kept for backwards compatibility.


gwbres (approved):

gwbres:
I'll eventually conclude this TEQC like ascii plot, which will be very convenient for people coming from `teqc`. 
As for the SINEX crate, it's in standby until we make progress on PPP and DORIS

Issue #219, opened 2024-03-24
Panic with `invalid Gregorian date: Carry` while position-solving


jigpu:
I've been experimenting with the position solving function of rinex-cli and have bumped into a problem when processing some Rinex files. In particular, I get the following panic from Rust:

~~~
thread 'main' panicked at ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hifitime-3.9.0/src/epoch.rs:827:14:
invalid Gregorian date: Carry
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: core::result::Result<T,E>::expect
             at /usr/src/debug/rust/rustc-1.76.0-src/library/core/src/result.rs:1030:23
   4: hifitime::epoch::Epoch::from_gregorian_utc
             at ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hifitime-3.9.0/src/epoch.rs:826:9
   5: rinex::epoch::parse_in_timescale
             at ./rinex/src/epoch/mod.rs:228:25
   6: rinex::epoch::parse_utc
             at ./rinex/src/epoch/mod.rs:254:5
   7: rinex::observation::record::is_new_epoch
             at ./rinex/src/observation/record.rs:144:13
   8: rinex::record::is_new_epoch
             at ./rinex/src/record.rs:264:34
   9: rinex::record::parse_record
             at ./rinex/src/record.rs:396:29
  10: rinex::Rinex::from_path
             at ./rinex/src/lib.rs:811:34
  11: rinex_cli::user_data_parsing
             at ./rinex-cli/src/main.rs:110:28
  12: rinex_cli::main
             at ./rinex-cli/src/main.rs:156:24
  13: core::ops::function::FnOnce::call_once
             at /usr/src/debug/rust/rustc-1.76.0-src/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
~~~

### Reproduction Steps
1. Download [mcso082.zip](https://github.com/georust/rinex/files/14736553/mcso082.zip) (source: [https://www.ngs.noaa.gov/UFCORS/](UFCORS))
2. Unzip to a temporary directory
3. Run rinex-cli as follows: `./target/debug/rinex-cli -f tmp/mcso0820.24o -f tmp/mcso0820.24n -f tmp/igr23065.sp3 positioning --spp`

### Notes
[...]

PR #222, opened 2024-03-28
Fix clock offset scaling, especially in sp3 interpolation

gwbres:
- [x] Solves #219


lnicola:
Ha! You can merge without a review! :grinning:

gwbres:
> Ha! You can merge without a review! 😀

Yeah, sorry but I was pretty sure about this one, and we had a very interesting pending discussion (see the Issue portal) that I wanted to update

lnicola:
It's fine, just that a while ago you couldn't merge without a review and couldn't self-approve either.

sp3/src/tests/parser_3d.rs
```diff
@@ -82,15 +82,15 @@ mod test {
         for (epoch, sv, clock) in sp3.sv_clock() {
             assert_eq!(epoch, Epoch::from_str("2019-10-27T00:00:00 GPST").unwrap());
             if sv == sv!("C01") {
-                assert_eq!(clock, 63.035497, "bad clock data");
+                assert!((clock - 63.035497E-6).abs() < 1E-9, "bad clock data");
```
lnicola, line 85: Not sure it's worth using it, but https://docs.rs/approx.

gwbres: Yeah I should look into adding this (only in test mode )

PR #224, opened 2024-03-29
Swap latitude and longitude in GPX output

jigpu:
The geo_point::Point is initalized with an (X, Y) coordinate, which it interprets as (longitude, latitude). We currently intialize the point in the wrong order which results in incorrect GPX output.


gwbres (approved):

gwbres:
@jigpu Thank you for your contribution !

PR #226, opened 2024-03-31
Handle (ignore) observation epochs containing event flags

jigpu:
The epochs in an observation file contain flags that can have a value from 0 - 6. If the value is in the range 2 - 5 then it is an "event flag" that signals the following lines will contain type-specific data. Epochs with these particular flags are allowed to not have a date (if it is "not significant"), which would trip up our parsing of the file.

This patchset updates the parsing to better handle these situations. We don't yet *do* anything with these events, but not falling on our face is still an improvement :)

While we're at it, fix a bug in the meterological code that would trigger a test failure in the presence of some of the other patches in this set.

Fixes: https://github.com/georust/rinex/issues/219


gwbres:
Thank you for your contributions! 
I gave it a test and everything works fine.
Moving the Epoch Flag since it is totally dedicated to Observation RINEX, is a good idea.

Planning: i'm working on positioning /performance upgrades in the rtk-0.5 branch, I will publish RINEX 0.16 once this one gets merged. If you're interested in an intermediate release, I can do that just let me know

gwbres (approved):

PR #230, opened 2024-04-13
Improve performance

gwbres:
This branch improves overal performances severely, especially in PPP opmode.
By improving the way we interface to the RTK core, we reduce the processing time severily.
PPP resolution of `test_resources/ESBCDNK-2020` (24hr) is reduced from 2'30 to 3.5sec on my computer.

NB: this will remain open until we're 100% sure data precision is preserved.


rinex-cli/src/positioning/interp/time.rs
```diff
@@ -0,0 +1,220 @@
...
+        if let Some(buf) = self.buffers.get_mut(&sv) {
+            buf.fill((t, dt));
+        } else {
+            let mut buf = Buffer::malloc(3);
+            buf.fill((t, dt));
+            self.buffers.insert(sv, buf);
+        }
+        self.ptr = Some(t);
+    }
+    // Consumses n epoch
```
lnicola, line 82:
```suggestion
    // Consumes n epochs
```

rinex-cli/src/positioning/interp/time.rs
```diff
@@ -0,0 +1,220 @@
...
+use gnss_rtk::prelude::{Epoch, SV};
+use std::collections::HashMap;
+
+use super::Buffer as BufferTrait;
+
+struct Buffer {
+    inner: Vec<(Epoch, f64)>, //TODO
+}
+
+impl BufferTrait<f64> for Buffer {
```
lnicola, line 11: I don't think the trait actually makes this code clearer, a wrapper over `Vec<f64>` (or `[f64; 3]' or `StaticVec<f64; 3>`) might be sufficient.

rinex-cli/src/positioning/interp/time.rs
```diff
@@ -0,0 +1,220 @@
...
+            buf.fill((t, dt));
+        } else {
+            let mut buf = Buffer::malloc(3);
+            buf.fill((t, dt));
+            self.buffers.insert(sv, buf);
+        }
+        self.ptr = Some(t);
+    }
+    // Consumses n epoch
+    fn consume(&mut self, total: usize) {
```
lnicola, line 83: I can't easily follow this, is it a push-based https://doc.rust-lang.org/std/vec/struct.Vec.html#method.chunks_exact?

rinex-cli/src/graph/record/observation.rs
```diff
@@ -1,95 +1,151 @@
...
-        for (sv, observations) in vehicles {
-            for (observable, data) in observations {
-                let observable_code = observable.to_string();
-                let physics = observable_to_physics(observable);
-                let y = data.obs;
-                let cycle_slip = match data.lli {
-                    Some(lli) => lli.intersects(LliFlags::LOCK_LOSS),
-                    _ => false,
-                };
+                    if let Some(data) = dataset_good.get_mut(&physics) {
```
lnicola, line 85: Maybe https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry?

rinex-cli/src/positioning/mod.rs
```diff
@@ -206,52 +211,21 @@ pub fn precise_positioning(ctx: &Context, matches: &ArgMatches) -> Result<(), Er
         panic!("High precision orbits (SP3) are unfortunately mandatory at the moment..");
     }
 
+    let mut orbit = RefCell::new(OrbitInterpolator::from_ctx(
```
lnicola, line 214: Is the `RefCell` here necessary?

lnicola:
For the record, I can't build this:

```
    Checking rstats v2.0.1
error: no rules expected the token `"point being added is coincident with gm"`
   --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecg.rs:447:43
    |
447 |             return re_error("arith",here!("point being added is coincident with gm"))?;
    |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
    |
    = note: while trying to match end of macro

error: no rules expected the token `"point being removed is coincident with gm"`
   --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecg.rs:458:43
    |
458 |             return re_error("arith",here!("point being removed is coincident with gm"))?; 
    |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
    |
    = note: while trying to match end of macro

error: no rules expected the token `"point being removed is coincident with gm"`
   --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecg.rs:469:43
    |
469 |             return re_error("arith",here!("point being removed is coincident with gm"))?; 
    |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
    |
    = note: while trying to match end of macro

error: no rules expected the token `"radius: invalid subscript"`
   --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecvec.rs:264:47
    |
264 |             return re_error("DataError",here!("radius: invalid subscript"))?;
    |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
    |
    = note: while trying to match end of macro

error: no rules expected the token `"sigvec given no data"`
   --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecvec.rs:303:43
    |
303 |             return re_error("empty",here!("sigvec given no data"))?;
    |                                           ^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
    |
    = note: while trying to match end of macro
```

rinex-cli/src/positioning/interp/time.rs
gwbres: It is pull based (if I follow you correctly), meaning that I pull one epoch at a time from the iterator

rinex-cli/src/positioning/mod.rs
gwbres: It is how I solved the following problem (haven't gone through your suggestions yet).
The RTK core uses function pointers (`Fn` closures) at instant `t` when it need some information to proceed.
It is a simple yet efficient method to make RTK agnostic of the end user (RINEX in our case).
One example of that is `Fn (t, sv)` to request results of position interpolation.
The "problem" is that Rust is not smart enough to understand this application is not multithreaded at the moment, and we're not allowed to capture the RinexContext (the data source, data provider, in our case) without troubles..

gwbres:
> For the record, I can't build this:
> 
> ```
>     Checking rstats v2.0.1
> error: no rules expected the token `"point being added is coincident with gm"`
>    --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecg.rs:447:43
>     |
> 447 |             return re_error("arith",here!("point being added is coincident with gm"))?;
>     |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
>     |
>     = note: while trying to match end of macro
> 
> error: no rules expected the token `"point being removed is coincident with gm"`
>    --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecg.rs:458:43
>     |
> 458 |             return re_error("arith",here!("point being removed is coincident with gm"))?; 
>     |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
>     |
>     = note: while trying to match end of macro
> 
> error: no rules expected the token `"point being removed is coincident with gm"`
>    --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecg.rs:469:43
>     |
> 469 |             return re_error("arith",here!("point being removed is coincident with gm"))?; 
>     |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
>     |
>     = note: while trying to match end of macro
> 
> error: no rules expected the token `"radius: invalid subscript"`
>    --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecvec.rs:264:47
>     |
> 264 |             return re_error("DataError",here!("radius: invalid subscript"))?;
>     |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
>     |
>     = note: while trying to match end of macro
> 
> error: no rules expected the token `"sigvec given no data"`
>    --> /home/grayshade/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstats-2.0.1/src/vecvec.rs:303:43
>     |
> 303 |             return re_error("empty",here!("sigvec given no data"))?;
>     |                                           ^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
>     |
>     = note: while trying to match end of macro
> ```

The Georust RINEX toolsuite does pass on my side and on the github worker.
But I do face the same problem when compiling the RTK core by itself, which is almost unbearable.
Therefore I always pass by the RINEX toolsuite, which is my main entry point anyways.
This appeared about last december. 

@ChristopherRabotin 

[rstat is a dependency](https://github.com/nyx-space/nyx/blob/master/Cargo.toml#L63) that Nyx has to get rid of. Have no idea why it behaves so strange nor what it does

rinex-cli/src/positioning/interp/time.rs
gwbres: I was interested in having this Trait because we need both spatial and time interpolation in this topic, therefore they share the same logic. Especially how we manage in the internal fifo against data gaps.

Note that data gaps are not correctly handled in the current implementation. This proposed form is a first solution, probably too stringent (does not tolerate any gaps, since we reset the fifo when a gap appears), but we will improve that in the future (simple detail)

ChristopherRabotin:
Oof, yeah, I often have issues with rstats. The crate is dependent on
another one which is maintained by the same person, and he'll update the
dependency but only make a patch version change but it should be a minor
version number change.

Can you try to run "cargo update"? That usually solves the issue for me.

lnicola:
> Can you try to run "cargo update"? That usually solves the issue for me.

Thanks, that worked. I only tried `cargo update -p rstats`, but that `here!` macro is from.. `indxvec`, which is the scariest crate name I've ever seen :roll_eyes:.

gwbres:
@ChristopherRabotin, @lnicola 

thank you guys !

rinex-cli/src/positioning/interp/time.rs
gwbres: I don't see how we could use methods like chunk_exact because we have no idea how many times we need to pull to complete at least one epoch. One Epoch is made of several SV and this number varies with time, as SV go out of sight

rinex-cli/src/positioning/mod.rs
lnicola: > The RTK core uses function pointers (Fn closures) at instant t when it needs some information.

I see what you mean. I didn't test, but `gnss-rtk` might be able to take a `FnMut` instead of `Fn`. `resolve` takes `&mut self`, so that might work.

rinex-cli/src/positioning/interp/time.rs
lnicola: I'm missing the larger context, but a `struct Buffer(Vec<f64>);` would work just as well and be easier to follow. The trait has a single implementation right now, and it seems unlikely it will get others, but I might be wrong.

rinex-cli/src/positioning/interp/orbit.rs
```diff
@@ -0,0 +1,398 @@
...
+    }
+    fn get(&self, idx: usize) -> Option<&(Epoch, (f64, f64, f64))> {
+        self.inner.get(idx)
+    }
+    fn latest(&self) -> Option<&(Epoch, (f64, f64, f64))> {
+        self.inner.last()
+    }
+    fn push(&mut self, x_j: (Epoch, (f64, f64, f64))) {
+        if self.len() == self.order + 1 {
+            let mut index = 0;
```
lnicola, line 43: Is this `self.inner.remove(0)`?

rinex-cli/src/positioning/interp/orbit.rs
```diff
@@ -0,0 +1,398 @@
...
+            order,
+            inner: Vec::with_capacity(order + 1),
+        }
+    }
+    fn len(&self) -> usize {
+        self.inner.len()
+    }
+    /// Fill internal Buffer, which does not tolerate data gaps
+    fn at_capacity(&self) -> bool {
+        self.inner.len() == self.order + 1
```
lnicola, line 30: Ooh, this makes more sense than comparing against 3 like the previous version did :upside_down_face:.

rinex-cli/src/positioning/interp/orbit.rs
lnicola: By the way, this is going to be the least of your problems, but you can use a ring buffer to store the samples, i.e. keep track of the number of samples and index of the last one.

rinex-cli/src/positioning/interp/orbit.rs
```diff
@@ -0,0 +1,398 @@
...
+        if let Some(buf) = self.buffers.get_mut(&sv) {
+            buf.fill((t, pos));
+        } else {
+            let mut buf = Buffer::malloc(self.order);
+            buf.fill((t, pos));
+            self.buffers.insert(sv, buf);
+        }
+        self.ptr = Some(t);
+    }
+    // Consumses n epoch
```
lnicola, line 92:
```suggestion
    // Consumes n epochs
```

rinex-cli/src/positioning/mod.rs
gwbres: Interesting, I did not know of FnMut

rinex-cli/src/positioning/interp/orbit.rs
gwbres: I simplified the logic a bit, removed the trait and made sure it still works.
This behaves like a FIFO of a strict depth

rinex-cli/src/positioning/interp/orbit.rs
gwbres: There are two dfiferent implementations of the Trait at the moment.
Orbital implementation is where this may vary

lnicola: No, there are two identical implementations of the trait:

image

You could do:

```rust
struct RingBuffer<T> {
    inner: Box<[T]>, // or Vec<T>
    head: usize,
    tail: usize,
}

type TimeBuffer = RingBuffer<(Epoch, f64)>;
type PositionBuffer = RingBuffer<(Epoch, (f64, f64, f64))>;
```

PR #238, opened 2024-04-24
Bugfix and refactoring in IonMessage parsing.

fedosgad:
Note: I only implemented 4 field parsing and not arbitrary field count parsing because:
1. 4 field parsing is used most often, other parameter counts are used only once
2. I don't know a simple way to do so (feel free to correct)

### DIFF
--- a/rinex-cli/src/graph/record/ionosphere.rs
+++ b/rinex-cli/src/graph/record/ionosphere.rs
@@ -61,7 +61,7 @@ pub fn plot_ionospheric_delay(ctx: &RnxContext, plot_ctx: &mut PlotContext) {
                             let (lat, lon) = lat_lon_ddeg;
                             let freq = Carrier::from_observable(sv.constellation, code).ok()?;
                             let ionod_corr =
-                                nav.ionod_correction(*t, elev, azim, lat, lon, freq)?;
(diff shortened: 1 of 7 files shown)


gwbres:
Hello, 

sounds like we can then take advantage of it when actually parsing the file.
I will look into this tonight, will keep you updated

gwbres:
Hello, 

I added code in src/Header to handle the (very) old `ION ALPHA` and `ION BETA` which would work like RINEX3, and are either RINEX1 or 2. 

It also turns out the current implementation is wrong because we only support a single Klobuchar model to be described in the header. I corrected this to work with real RINEX 2 and 3, which is possibly several models (per constellation) and possibly NG Model as well

fedosgad:
I submitted #239 - it contains v2 parsing code. It is not compatible with this PR unfortunately, so you can close it if you want, but I hope code may be helpful.

gwbres:
> I submitted #239 - it contains v2 parsing code. It is not compatible with this PR unfortunately, so you can close it if you want, but I hope code may be helpful.

Ok, I'll just merge the few bits that this one improves, that allows to have several corrections - per constellation, which may happen in RINEX3.

Also, thank you for taking the time to write a test

PR #239, opened 2024-04-25
RINEX v2 ionosphere parameter parsing

fedosgad:

PR #241, opened 2024-04-29
Fix incorrect RINEX v2 ionosphere parameter parsing

fedosgad:
In #239 I made a mistake in parsing code for `IONO ...` header strings: 
String format is `2X,4D12.4` (https://files.igs.org/pub/data/format/rinex211.txt, table A3), which means there must 2 empty characters at the beginning of the line. In my code, I skip first 4 characters instead of 2, which can lead to errors in `f64::from_str` further down the line. Unfortunately, existing tests were incorrect too, so mistake hasn't been caught early.

I fixed parsing code and added more tests from credible sources.
Sorry for the bug.


gwbres:
Hello, no worries ! it went unoticed for the simple reason we don't have such test data

gwbres (approved):

gwbres:
I'm thinking we should maybe host a complete PPP set in V2, like we do in V3, that would be the best thing to do to test all functionalities

PR #249, opened 2024-06-08
Improve support for BDRC ephemeris

FlowLoveV:
-  Fix bugs in kepler orbit calculation.Correct calculation of GPS,BDS,GALIEO satellite position and velocity is now supported on all versions of rinex.QZSS support has not been tested.
- The next step is to refer to rtklib and update the satellite ephemeris calculations for GLONASS and SBAS to get their exact positions instead of the approximate positions reported.
-  Corrected BDS ephemeris update time to one hour

PR #255, opened 2024-07-21
Switch from MetaFile to embedded Almanac ephem

ChristopherRabotin:
Also _temporarily_ switch to my fork of gnss-rtk until the PR is merged.

Companion PR to https://github.com/rtk-rs/gnss-rtk/pull/18 .

### DIFF
--- a/rinex-cli/Cargo.toml
+++ b/rinex-cli/Cargo.toml
@@ -48,9 +48,7 @@ serde = { version = "1.0", default-features = false, features = ["derive"] }
 
 # gnss-rtk = { version = "0.4.5", features = ["serde"] }
 # gnss-rtk = { path = "../../gnss-rtk", features = ["serde"] }
-gnss-rtk = { git = "https://github.com/ChristopherRabotin/gnss-rtk", branch = "fix/anise-frame-data", features = [
-    "serde",
-] } # TODO: Revert to official git after merging gnss-rtk change.
+gnss-rtk = { git = "https://github.com/rtk-rs/gnss-rtk", branch = "main", features = ["serde"] }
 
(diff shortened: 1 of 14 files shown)


gwbres:
Ok so this is the last topic to solve. 
I rebased everything. 
Your data analysis was correct: the modified quantities and scaling are mandatory (I kept the comment in the code).

The performance is degraded by about 10x/100x times I would say, in this contribution.
There is probably still a tiny error in what we provide to `try_keplerian()`.
Right now, we're getting errors of about 100/500m while we used to achieve 10m.

The error is purely located in this topic, and timing is correct, so the error is solely in the SV state vector.
You can give it a try by using radio based navigation :

```bash
./target/release/rinex-cli \ 
   -f test_resources/CRNX/V3/ESBC00DNK_R_20201770000_01D_30S_MO.crx.gz \
   -f test_resources/NAV/V3/ESBC00DNK_R_20201770000_01D_MN.rnx.gz \
   -P GPS \ 
   -p
```

One other problem (might be solved in two steps), si that GPS/GAL do not exhibit the same performance as BDS (switch the filter in previous command). The state of the art performance was obtained by this contribution : 
https://github.com/georust/rinex/pull/249 and we should refer to its navigation:ephemeris content.

gwbres:
Unfortunately, it is impossible to deploy the solver in rinex:main in the current setup, and it has been like this for too long.
So I'm merging this proposal as is, although it is not 100 % correct, at least it fixes several issues. 
I will copy and paste the previous message, in the nyx:discussion, so we keep investigating the issue

gwbres (approved):

PR #274, opened 2024-10-10
BINEX

gwbres:
* introducing BINEX parser


gwbres:
The only official BINEX specs are https://www.unavco.org/data/gps-gnss/data-formats/binex/binex.html

The first objective is to achieve one first test method that decodes one message correctly.

The ultimate objective is to be able to provide seamless interface between RINEX and BINEX (ideally both ways)

binex/src/decoder.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+impl<R: Read> Iterator for Decoder<R> {
+    type Item = Result<Message, Error>;
+    /// Parse next message in stream
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.reader.read(&mut self.buffer) {
+            Ok(size) => {},
+            Err(e) => {
```
lnicola, line 55: Shouldn't this propagate the error (`self.reader.read(&mut self.buffer)?`)? When the end of the stream is reached, `read` returns 0, not an error.

binex/src/decoder.rs
```diff
@@ -0,0 +1,64 @@
...
+                return None; // EOS
+            },
+        }
+        match Message::decode(&self.buffer) {
+            Ok(msg) => Some(Ok(msg)),
+            Err(e) => None,
```
lnicola, line 61:
```suggestion
            Err(e) => Some(e),
```

?

binex/src/decoder.rs
gwbres: Yes it needs to propage at least the EOS error (and possibly other major errors), because that's the only way for the user to determine that the EOS is reached.

I'm currently working on the actual message decoding and am not focused on the Decoder (final layer)

gwbres:
@lnicola,

I would like to make this `no-std` compatible, maybe not right now, but in the long term I think it definitely should:

* this is just a simple parser and forger
* forging is data production oriented and you want to use this nearby the gnss firmware or on the same plateform

do you have recommendations on how to replace `Vec` in this scenario ? maybe a small external library so we don't have too much code to write. It is most convenient to use Vec in the decoder Buffer currently, but it will then have to change

binex/src/decoder.rs
gwbres: It's starting to work, I'm able to fill the buffer (I assigned an arbitrary size of 1024), 128 should be minimum.
For some reason, I can only get the I/O read interface to work, if I allocate my Vector with `vec = [0; 1024]`, which seems to _actually_ fill the buffer. At first I used `Vec::with_capacity(1024)`, which does not seem to fill the buffer, `.read()` then presumes the buffer is actually empty :/

binex/src/decoder.rs
gwbres: I slightly modified the architecture.
See now line 82 in `decoder.rs` is the exact instant the internal buffer is in good condition to attempt parsing.
3 cases:

1. if parsing is successful, increment the pointer by the message length 
2. if the special SOF marker is not found, the entire buffer is rubbish and we can directly proceed to a new read
3. if parsing is in failure, most likely due to unsupported format, I simply increment by 1 byte.
They're might be something smarter to do here. And we keep attempt parsing the remaining bytes

lnicola:
Sorry, I saw the notification after you merged the PR, then I forgot to answer. If the format has some built-in limits (say, for the number of frames in a record, the maximum size of a record etc.) then something like `arrayvec` should work work. If the records can be arbitrarily large, then I don't think that's going to be compatible with `#[no_std]` without using the `alloc` crate.

gwbres:
> Sorry, I saw the notification after you merged the PR, then I forgot to answer

No worries, your input is always much appreciated. The no-std support can be a long term thing, it does not have to be accomplished right away, I personnaly have very limited experience with that. I think it's an important topic, because open-source GNSS firmware (embedded) could then easily use the BINEX encoder/decoder

>  If the format has some built-in limits (say, for the number of frames in a record, the maximum size of a record etc.) then something like `arrayvec` should work

unfortunately not, you cannot say at the beginning of the parsing process, how many bytes you will encounter for the Geodetic and Signal frames. For the first one (comment/description) you can have as many comments as needed: I currently use a `Vec<>` (convenient but then incompatible). For the latter, it depends on the GNSS receiver capabilities, and it may be able to sample many carriers so you then have to describe all your observations (not supported yet).

> If the records can be arbitrarily large, then I don't think that's going to be compatible with `#[no_std]` without using the `alloc` crate

Yep I saw a couple of times the possibility to implement no-std with an allocator, but I need to study how that works

lnicola:
> unfortunately not, you cannot say at the beginning of the parsing process, how many bytes you will encounter for the Geodetic and Signal frames

No 64 KB record limit? No 16-bit length field? :-)

> Yep I saw a couple of times the possibility to implement no-std with an allocator, but I need to study how that works

I think you can use `alloc::Vec<T>` and let the user figure out where to get an allocator from.

gwbres:
> No 64 KB record limit? No 16-bit length field? :-)

Hum, at first I thought that was correct (have not read the specs since last week 😄), but unfortunately that does not stand in the case of the BINEX protocol.

In BINEX, a few sensitive bytes are encoded using the so called [1-4 BNXI algorithm](https://github.com/georust/rinex/blob/main/binex/src/message/mod.rs#L373), which as the capacity to describe one integer (usually U32) using between 1 up to 4 bytes, depending on the value to be described. [That applies to the message length (MLEN)](https://github.com/georust/rinex/blob/main/binex/src/message/mod.rs#L144) and gives up to 2^32 bytes in a hypothetical record

gwbres:
@lnicola

> No 64 KB record limit? No 16-bit length field? :-)
> I think you can use `alloc::Vec<T>` and let the user figure out where to get an allocator from.
>> In BINEX, a few sensitive bytes are encoded using the so called 
>> [1-4 BNXI algorithm](https://github.com/georust/rinex/blob/main/binex/src/message/mod.rs#L373), 
>> which can describe an integer value (usually U32) using only the required amount of bytes
>>  to describe the actual value. [That applies to the message 
>> length (MLEN)](https://github.com/georust/rinex/blob/main/binex/src/message/mod.rs#L144) and gives up to 2^32 hypothetical record length

correcting this statement which is wong or at least inacurate. It is true that a record may hypothetically be 2^27 byte long using this protocol, but it only applies to prototyping frames (closed source and not official frames) and all we need is someway to describe partial / incomplete frames. It does not mean we cannot operate using a static buffer with reasannable size. 

I'm still in the process of learning this protocol

PR #283, opened 2024-12-23
Fix CGGTTS error when the observation interval is 1 second

vicalloy:
I tried to use rinex-cli to create CGGTTS from RINEX, but it failed(not enough candidates match post-fit criteria). After some debugging, I found that the issue occurs when the observation interval is 1 second.

When the observation interval is 1 second, the line if t >= next_release is called twice within the same track, causing the track to be reset twice. (Note: t == next_release, but sched.next_track_start(*t) does not update next_release.)

I attempted to fix the bug, and the code now works. However, I am not sure if the measurements in the track are correct. Please review the code.

Thank you.

### DIFF
--- a/rinex-cli/src/positioning/cggtts/mod.rs
+++ b/rinex-cli/src/positioning/cggtts/mod.rs
@@ -417,11 +417,8 @@ pub fn resolve<'a, 'b, CK: ClockStateProvider, O: OrbitSource>(
         next_release = Some(sched.next_track_start(*t));
         // If the last epoch is reached, should release track
         let next_release_duration = next_release.unwrap() - *t;
-        should_release = (
-            next_release_duration <= dominant_sampling_period
-        ) && (
-            next_release_duration > Duration::ZERO
-        );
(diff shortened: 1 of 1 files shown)


gwbres:
Thank you for your contribution 👍  

I just tested it with `tutorials/CGGTTS` and it works as previously.
I'm inquiring "fast" RINex so we can actually test that (I personnaly have never done that before)

- we have  `2024-DOY=127` with `OBS/V3/240506_glacier.gz`.
Apparently the CGGTTS tracker uses `RINEx.dominant_sample_rate()` as a starting point, which seems reasaonnable.
But for some reason, this method is in failure on that particular dataset. I will inquire that as well.

I am not surprized the current logic does not work in "extreme" cases. First it has not been tested in such conditions. Secondly, it is too complex and should be improved.

The tracking logic is being improved in the `0.17-rc` branch that I hope to merge in the next two weeks, so stay tuned. It will come with a lot of API improvements, simplifications and architecture improvements

Issue #285, opened 2024-12-24
The tracking schedule and trk_midpoint are not correct in the CGGTTS Conventional Observation schedule.


vicalloy:
According to [CGGTTS-Version 2E](https://www.bipm.org/documents/20126/52718503/G1-2015.pdf/f49995a3-970b-a6a5-9124-cc0568f85450)

1. `A minimum of 3 minutes was then foreseen in the BIPM schedule to give to the receiver the time to track the following scheduled satellite.`   
2. `The data to be used are the pseudorange measurements collected during the 13 minutes starting with the date specified in the BIPM tracking schedule.`

If the Common-View Duration is 960 seconds, the `trk_duration` should be 780 seconds (13 minutes * 60 seconds). The `trk_midpoint` should be calculated as `next_release - 3 minutes - (780 seconds / 2)`.

https://github.com/georust/rinex/blob/fc20d8c178e08a235a7430fc5a49d9c204a5b3bd/rinex-cli/src/positioning/cggtts/mod.rs#L422

vicalloy:
Thank you for your work, but changing only the `trk_midpoint` is not sufficient.  
The CV period is 16 minutes, which includes 3 minutes of preparation time and 13 minutes of tracking time. We should skip the 3-minute preparation time.  The `TRKL` in CGGTTS should be `780`, not `960`.

I tried to create a pull request(https://github.com/georust/rinex/pull/288) to fix this issue.  
I tested it with my data, and it works. However, some variable names may still seem odd, which is why I didn’t fix  them initially.
- `cggtts::track::Scheduler`:  
    - `trk_duration` should be renamed to `cv_duration`.  (should keep `trk_duration`?)
    - `bipm_tracking_duration` should be renamed to `bipm_cv_duration`.  
- `cv_duration` in CGGTTS is `TRKL`.

```
                  tracking start point                           
                    │                                            
      prepare(3min) │                                            
                    │  tracking(13mins)         │                
      │         │◄──┘                           │                
 ─────┼─────────┴───────────────────────────────┼────────        
      │                                         │                
             cv period(16min）                                    
      ▲                                         ▲                
      │                                         │                
    cv start point                       cv/tracking end point   
```

gwbres:
Hello @vicalloy, to keep you updated:

* the upgraded rinex backend is merged into the master branch, in the `rinex-backend` branch.
* In this branch, the applications are upgraded to the new internal library
*  the PPP is updated and it works fine
[...]

PR #287, opened 2024-12-26
Fix track midpoint determination, in CGGTTS tracking

gwbres:
- [x] #285 

To test the tracker, I used `./tutorials/CGGTTS/mojndnk` focused on `E05+E09`

Results with current version:

image

Results with the proposal :

image

@vicalloy,

note that the CGGTTS scheduler is tested in cggtts/track/scheduler, for example the initial point in any day is tested thoroughly

image


gwbres:
@vicalloy ,

we're locked to `nyx` and `rtk-rs` stable branches, don't forge to run a quick `cargo update` to make sure you collect their latest state

PR #289, opened 2025-01-06
Fix CGGTTS, reset the tracker after it is released.

vicalloy:
If `tracker.fit fails`, `tracker.reset()` will not be called, and the samples will be kept for the next period.
This commit resets the tracker after tacker release.

### DIFF
--- a/rinex-cli/src/positioning/cggtts/mod.rs
+++ b/rinex-cli/src/positioning/cggtts/mod.rs
@@ -256,6 +256,16 @@ pub fn resolve<'a, 'b, CK: ClockStateProvider, O: OrbitSource>(
 
                     // TODO: RTK
                     //candidate.set_remote_observations(remote);
+                    let target = &(*sv, observable.clone());
+
+                    let tracker = match trackers.get_mut(target) {
+                        None => {
+                            // initialize new tracker
+                            trackers.insert((*sv, observable.clone()), SVTracker::default());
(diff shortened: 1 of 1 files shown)


gwbres (approved):

gwbres:
my big question in CGGTTS track fitting process, is whether we should tolerate data gaps or not.
I presume we should, because it is unrealistic to hope for 15' steady observations.
And after all, that is the whole point of a  _fitting_ algorithm

vicalloy:
> my big question in CGGTTS track fitting process, is whether we should tolerate data gaps or not. I presume we should, because it is unrealistic to hope for 15' steady observations. And after all, that is the whole point of a _fitting_ algorithm

Yes, I have the same question. In my scenario, my sample interval is 1 second, and missing one sample is not a significant issue for the result. 
Maybe we can add a ratio parameter for the minimum sample count.

gwbres:
The tracker should simply have a gap `Duration` tolerance, which seems the best parameter to tune by the user (depending on requirements and application). I would also argue that the tracker's default behavior should be very tolerant (tolerate almost "any" gap) for very basic applications. To me that would be 3 points per CvPeriod. Which is incompatible with high quality CGGTTS, but just works. In your 1Hz (""fast"" observation application), you could for example specify a 30s tolerance, which would be compatible with high quality CGGTTTS

PR #346, opened 2025-04-20
added support for D notation parsing of Ionospheric Corrections in V3…

segrote:
Some V3 RINEX files use "D" exponential notation which is mostly supported aside from the parsing of IONOSPHERIC CORR parameters. This PR aims to support such RINEX files that may be sourced from IGS stations like: KERG00ATF (shown below)

image


gwbres:
Thank you for your contribution. I did not pay attention that detail once again !
We will release a new version as soon as a new Hifitime version is published. Stay tuned

PR #411, opened 2026-04-15
numdiff: document historical limit correctly

Artoria2e5:
numdiff: document Hatanaka correctly
* Hatanaka is lossless. Did an LLM hallucinate this?
* Although rnx2crx uses `m` = 3, the accompanying crx2rnx program uses
  `M`` = 5 as the maximum level.
  This was explicitly mentioned in Hatanaka (2008) as a point of possible
  flexibility.
* The compression "expert" is not making use of a larger `m` either.
* A larger `m` is not necessarily better, see Hatanaka tables 3 & 4.
* Cite Hatanaka correctly.

### DIFF
--- a/src/hatanaka/textdiff.rs
+++ b/src/hatanaka/textdiff.rs
@@ -1,11 +1,13 @@
 //! Y. Hatanaka lossless TextDiff algorithm
 
-/// [TextDiff] is a structure that implements the Text diff. algorithm
-/// designed by Y. Hatanaka, which is a lossless text compression algorithm.
+/// [TextDiff] implements Hatakana's text differencing scheme,
+/// which improves compression of text data.
+///
 /// [TextDiff] in its current form does not allow compressing & decompressing (back & forth)
 /// at the same time: you need two dedicated objects.
-/// This does not bother our application because it only operates in one way,
-/// but it is one aspect to keep in mind.
+///
+/// This is not a limitation to the CRX2RNX decompression or RNX2CRNX compression
+/// algorithm, because they only work one way.
 #[derive(Debug)]
 pub struct TextDiff {
     buffer: String,
--- a/src/hatanaka/decompressor/io.rs
+++ b/src/hatanaka/decompressor/io.rs
@@ -7,11 +7,13 @@ use crate::{
 
 use std::io::{BufRead, BufReader, Lines, Read};
 
-/// [DecompressorIO] is a [Decompressor] implementation that works directly
-/// on any [Read]ble I/O interface. Use it if your application to decompress
-/// CRINEX to parsable RINEX efficiently. Refer to [Decompressor] for its
-/// limitations. It implements the same internal parameters as the historical
-/// CRX2RNX tool.
+/// [DecompressorIO] emplements [DecompressorExpert] internally and works
+/// directly on any [Read]able I/O interface.
+///
+/// Use it if your application to decompress CRINEX to readable RINEX efficiently.
+/// Refer to [Decompressor] for more information on the algorithm and the limitations.
+/// The historical CRX2RNX tool uses M=5 as the maximal compression order to ever be supported:
+/// this implementation uses the same setup.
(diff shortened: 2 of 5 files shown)


Artoria2e5:
The sentence about it being lossy is also false. It is plain delta. It is lossless.

src/hatanaka/textdiff.rs
```diff
@@ -1,7 +1,7 @@
 //! Y. Hatanaka lossless TextDiff algorithm
 
-/// [TextDiff] is a structure that implements the Text diff. algorithm
-/// designed by Y. Hatanaka, which is a lossless text compression algorithm.
+/// [TextDiff] implements Hatakana's text differencing scheme,
+/// which supposedly improves compression of text data.
```
Artoria2e5: It is honestly dubious whether this scheme provides any benefit over no compression when used in the recommended scheme of "combine with a general-purpose compressor". Something dumb like gzip, maybe, but bzip2? xz?

Artoria2e5: > All comments must be resolved.

Uh oh

gwbres:
Thank you for the corrections, I will probably have more comments I will come back to it later, mostly when you are refering to the compression program, it is not clear to me whether you are refering to the historical program, or this library or the program based off this library.

It is true that the algorithm is lossless if correctly implemented (using integer numbers. For that Rust being strongly typed, it helped doing it correctly). But I am not sure it remains lossless on large compression orders. The purpose of the differentiation algorithm is to reduce the number of digits to encoder

> The compression "expert" is not making use of a larger m either.

that's not how it works, it allows to support larger `m`, m is dictated by the file content, as clarified by the new comments.

> A larger m is not necessarily better, see Hatanaka tables 3 & 4.

no one said it was

> Did an LLM hallucinate this?

there is no LLM involved either in this work or around here, nor people hallucinating

Artoria2e5:
> mostly when you are refering to the compression program, it is not clear to me whether you are refering to the historical program, or this library or the program based off this library.

"This program" refers to this library (crate rinex). The historical program is consistently referred to as crx2rnx or rnx2crx.

> there is no LLM involved either in this work or around here, nor people hallucinating

My bad. That was rude of me.

> It is true that the algorithm is lossless if correctly implemented ...  But I am not sure it remains lossless on large compression orders.

It remains lossless on large orders. Consider an induction argument: you know that going from 0 to 1 is lossless because you can just add them back. When you go from 1 to 2, you are basically doing the "0 to 1" operation to the order-1 list, and we already know that the "increment order by 1" operation is losslessly reversible. So the same goes for any positive integer order.

```
0: 1 4 9 16 25...
1: 1 3 5 7 9...
2: 1 2 2 2 2...
3: 1 1 1 1 1...
4: 1 0 0 0 0...
5: 1 0 0 0 0...
```

> no one said it was

Your comments keep saying that `m` (lowercase, i.e. level) = 5 is optimal when that does not seem to be the case.

> that's not how it works, it allows to support larger m, m is dictated by the file content, as clarified by the new comments.

The DE-compressor reads the lowercase `m` from file content. The compressor always has its own idea, see:

```rs
                            // first encounter: build kernel
                            let kernel = NumDiff::<M>::new(quantized, 3);
                            self.sv_kernels
                                .insert((signal.sv, signal.observable.clone()), kernel);

                            write!(w, "{}&{} ", 3, quantized)?;
```

You hard-code the level = 3.

gwbres:
> Your comments keep saying that m (lowercase, i.e. level) = 5 is optimal when that does not seem to be the case.

From the article level, 5 (as used by the historical tools) are said to be a good compromise

>  The DE-compressor reads the lowercase m from file content. The compressor always has its own idea, see:

that's correct, the decompressor need to decide which compression factor to use, and an interface to describe variations of that does not exist. The historical tool fixes M=5 to ever be supported, and seems to use m=3 all the time. So I simply recreated that behavior

gwbres (approved):

PR #421, opened 2026-05-28
Fixed bug in rinex 2 observation parsing that skipped SVs.

zzoip:
The bug could be seen in the tests
```rust
fn test_parse_v2_barq() {
        // extracted from v2/barq0.19d
        let content = " 19  3 12 16 36  0.0000000  0 15G08G10G14G16G18G20G22G26G27G32R04R05
        R06R19R20
111525030.92718  86902614.11057  21222508.060                    21222505.880
113917475.72718  88766858.51957  21677775.000                    21677773.380
117582228.43417  91622504.36956  22375154.360                    22375150.800
116496701.38017  90776637.01356  22168585.640                    22168581.380
120543799.60817  93930215.60156  22938722.560                    22938718.700
125411881.34917  97723533.46955  23865086.320                    23865082.480
124148906.10717  96739381.30556  23624751.240                    23624745.820
127085060.78216  99027334.17655  24183483.520                    24183486.540
105970568.06418  82574457.73858  20165528.740                    20165526.020
114286506.03418  89054414.91257  21747998.820                    21747997.860
113282224.20717  88108400.20417  21154656.900                    21154656.960
102507243.91718  79727864.12617  19176099.600                    19176101.880
116775184.83016                  21883618.780
112181738.38117  87252468.04516  20971192.900                    20971194.040
110923986.30317  86274222.72417  20743344.820                    20743348.200
";
        generic_observation_epoch_decoding_test(
            content,
            2,
            Constellation::Mixed,
            &[
                ("GPS", "L1,    L2,    C1,    P1,    P2,    P1,    P2"),
                ("GLO", "L1,    L2,    C1,    P1,    P2,    P1,    P2"),
            ],
            "2019-03-12T16:36:00 GPST",
            50,
            "2019-03-12T16:36:00 GPST",
            EpochFlag::Ok,
            None,
            vec![],
        );
    }
```

You can manually count 58 signals in the input, but the test was trickily passing by asserting with 50 signals. Basically there was a border case where some signal lines were ignored because obs_ptr was exceeding the amount of observables, causing the epoch to be finished missing the SVs that were remaining. Particularly for the test shown above, the last two lines were being ignored. Bug was fixed and the tests with wrong assertions were changed and are passing with correct expected values.

PR #422, opened 2026-05-31
Fixed errors when using feature "nav" due to hyperdual v1.5.0

zzoip:
Hyperdual 1.5.0 depends on nalgebra v0.34.0, thats why missmatched types errors appeared when using nav feature. The solution is to add hyperdual v1.4.0 to dependencies and use it in nav feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ obs = []
 nav = [
     "anise",
     "nalgebra",
-    "hyperdual"
+    "hyperdual",
 ]
 
 # Provides the special UT1-TAI methods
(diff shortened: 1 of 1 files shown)


gwbres (approved):

gwbres:
Hello,

are you sure you pushed everything you wanted ? This is all that I have on my side:
image

it should be declared in the dependencies with a specific version

sorry I did not get time to push a release the last few days, but I should have a little time tomorrow

zzoip:
Yup, that forces nav feature to use hyperdual 1.4.0 and with it, nalgebra 0.34

zzoip:
> it should be declared in the dependencies with a specific version

Sorry i just read this. The dependency is declared in the file already. There is this line:
```toml
hyperdual = { version = "=1.4.0", optional = true }
```
It was inserted by me in #421, now i added it to nav feature, thats why the only change is what you show in the screenshot.

Issue #426, opened 2026-07-14
i64 overflow while decoding a valid CRINEX V3 receiver clock differential


htyeim:
## Summary

`rinex 0.22.0` panics while decompressing the receiver-clock field of a CRINEX V3 observation file published by BKG.

The reference RNXCMP `CRX2RNX` implementation decodes the same input successfully.

## Input file

URL:

https://igs.bkg.bund.de/root_ftp/IGS/obs/2024/001/ACRG00GHA_R_20240010000_01D_30S_MO.crx.gz

Metadata:

- File size: `5,384,034` bytes
- SHA-256: `add6a7644f2600a8360a6e5050750d89892612e1a4b3583786476f44f3619386`
- CRINEX version: 3.0
- Embedded RINEX version: 3.05
- Created by RNX2CRX 4.1.0

## Minimal reproduction

`Cargo.toml`:

```toml
[package]
name = "rinex-crinex-overflow-repro"
version = "0.1.0"
edition = "2024"

[dependencies]
rinex = { version = "=0.22.0", features = ["obs"] }
```

```rust
use rinex::prelude::Rinex;

fn main() {
    let path = std::env::args()
        .nth(1)
        .expect("usage: repro <file.crx.gz>");

    let rinex = Rinex::from_gzip_file(path).unwrap();

    println!(
        "epochs={}",
        rinex
            .record
            .as_obs()
            .map_or(0, |observations| observations.len())
    );
}
```

Run:
```sh
curl -fL \
  -o sample.crx.gz \
  'https://igs.bkg.bund.de/root_ftp/IGS/obs/2024/001/ACRG00GHA_R_20240010000_01D_30S_MO.crx.gz'

RUST_BACKTRACE=1 cargo run -- sample.crx.gz
```
### Actual result

```sh
thread 'main' panicked at rinex-0.22.0/src/hatanaka/numdiff.rs:67:42:
attempt to multiply with overflow
```

<img width="1279" height="838" alt="Image" src="https://github.com/user-attachments/assets/ef0b42e2-3d1b-4586-b572-ad7a31e598e2" />

The relevant call chain is:
```rust
rinex::hatanaka::numdiff::NumDiff<5>::decompress
rinex::hatanaka::decompressor::DecompressorExpert<5>::run_clock
rinex::hatanaka::decompressor::Decompressor::decompress
rinex::Rinex::from_gzip_file
```

The apparent overflow site is:
[...]

PR #428, opened 2026-09-02
hatanaka: fix history buffer rotation in NumDiff

MrCry0:
## Summary

`NumDiff::rotate_history()` shifted only `buf[0..M-2]` into `buf[1..M-1]`,
leaving the last slot `buf[M-1]` untouched after construction. Any
`decompress()`/`compress()` call that reaches the maximal supported order
`M` (order 5 for the default `NumDiff<5>` used to decompress the CRINEX
clock offset field) then read that stale, never-updated slot instead of
the correct historical value.

Because Hatanaka differencing is stateful, the resulting error is not
self-correcting: it propagates into every following sample and grows
unbounded, eventually causing the reported `attempt to multiply with
overflow` panic deep into otherwise valid CRINEX V3 files that
legitimately reach that order.

The fix shifts `buf[0..M-1]` instead, so every slot of the history
buffer stays current.

## Out of scope

`NumDiff::compress()` has a separate, pre-existing limitation: it reads
one more history slot than `decompress()` for the same order, so
`compress()` at order `M` still indexes out of bounds. That is tracked
in \# 429 and intentionally not addressed here, since it does not affect
the decompression path reported in #426.

## Test plan

- Added `test_history_rotation_full_order`, a regression test that
  decompresses a synthetic order-5 sequence (ground truth computed
  independently of this crate) and checks the recovered values match.
  Confirmed it fails against the pre-fix code (values diverge, matching
  the divergence pattern that leads to the reported overflow) and passes
  after the fix.
- `cargo test -p rinex --lib hatanaka` passes.
- `cargo test -p rinex --lib` shows no new failures compared to `develop`
  (same pre-existing, unrelated `orbit_database_sanity` cascade failures
  in both cases).
- `rustfmt --check` on the changed file.

### DIFF
--- a/rinex/src/hatanaka/numdiff.rs
+++ b/rinex/src/hatanaka/numdiff.rs
@@ -48,7 +48,7 @@ impl<const M: usize> NumDiff<M> {
 
     /// Rotate internal buffer, take new sample into account.
     fn rotate_history(&mut self, data: i64) {
-        self.buf.copy_within(0..M - 2, 1);
+        self.buf.copy_within(0..M - 1, 1);
         self.buf[0] = data;
     }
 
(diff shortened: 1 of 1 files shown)


MrCry0:
This commit has been included to and merged with PR #430

Issue #429, opened 2026-09-02
hatanaka: NumDiff::compress indexes out of bounds at order M


MrCry0:
## Summary

`NumDiff<M>::compress()` panics with an index-out-of-bounds error when the
compression order reaches `M`, even though `M` is documented as the maximal
supported order and `NumDiff<M>::decompress()` handles order `M` fine.

Follow-up to #426 / #430, which fixed the history rotation but deliberately
left this separate limitation alone.

## Root cause

The two directions read a different number of history slots for the same
order `m`:

- `decompress()` computes the new value before rotating the buffer, so
  `buf[0]` is the previous sample and order `m` reads `buf[0..=m-1]`.
  This needs `M >= m`.
- `compress()` rotates first, so `buf[0]` is the new sample itself and
  order `m` reads `buf[0..=m]`. This needs `M >= m+1`.

So with the default `NumDiff<5>` (`M = 5`, buffer indices 0 to 4),
`compress()` at order 5 evaluates `self.buf[5]` and panics. The same
happens for `NumDiff<6>` at order 6 (`self.buf[6]`).

The `level > 6` guard in `new()` / `force_init()` does not catch this,
because it checks the order against the hard-coded ceiling of 6 rather
than against the buffer size `M`.

## Reproduction

```rust
use rinex::hatanaka::NumDiff;

let mut diff = NumDiff::<5>::new(126298057858, 5);
diff.compress(126282454570);
diff.compress(126267372371);
diff.compress(126252810509);
diff.compress(127814188268);
diff.compress(127800656941); // 5th call, order reaches 5
```

Output:

```
thread 'main' panicked at rinex/src/hatanaka/numdiff.rs:97:23:
index out of bounds: the len is 5 but the index is 5
```

## Reachability

`NumDiff` is re-exported from `rinex::hatanaka`, so any downstream user
following the type's own documentation ("M specifies the maximal
compression to ever be supported") can hit this. The in-tree `Compressor`
currently only uses `TextDiff` for the epoch line, so `rnx2crx` is not
affected today, but it will be once numerical observation compression is
wired in.

## Suggested fix

Either of:

1. Make `compress()` symmetric with `decompress()` by computing the
   difference before rotating, so both need `M >= m`. Existing
   `test_compression` values should be unchanged since they only use
   order 3 with `M = 6`.
2. Keep the current layout but validate `level` against `M` in `new()` /
   `force_init()` (`level + 1 <= M` for compression) and return an error
[...]

PR #430, opened 2026-09-02
hatanaka: make NumDiff::compress symmetric with decompress

### LINKED ISSUE #426: i64 overflow while decoding a valid CRINEX V3 receiver clock differential

MrCry0:
## Summary

Fixes #426 #429.

Stacked on #428: this branch contains that commit as well, since
compressing at order `M` needs the full history buffer that #428 repairs.
Only the top commit is new here. Merging #428 first will reduce this PR
to a single commit.

`NumDiff::compress()` rotated the history buffer *before* differencing,
so order `m` read `buf[0..=m]`, while `decompress()` computes first and
reads only `buf[0..m]`. A `NumDiff<M>` could therefore decompress at
order `M` but compressing at the same order indexed `buf[M]`, one past
the end of the buffer:

```
thread 'main' panicked at rinex/src/hatanaka/numdiff.rs:97:23:
index out of bounds: the len is 5 but the index is 5
```

The `level > 6` guard in `new()` / `force_init()` did not catch this
because it checks against the hard-coded ceiling of 6, not against `M`.

The fix computes the difference against the existing history and rotates
afterwards, mirroring `decompress()`. Both directions now read the same
slots for a given order, so a single `NumDiff<M>` supports orders up to
`M` for compression and decompression alike. Results are unchanged for
every order that previously worked.

## Test plan

- `test_compression_full_order`: compresses a 9-sample sequence at order 5
  with `NumDiff<5>` and checks the output against an independently
  computed reference stream. Panics with the index-out-of-bounds error on
  the old code, passes on the new code.
- `test_round_trip_full_order`: compress/decompress round trip with the
  same `NumDiff<M>` on both sides, for `M` in {3, 5, 6} and every order
  from 1 up to `M`, including a `force_init()` mid-stream. Panics on the
  old code as soon as the order reaches `M`, passes on the new code.
- The existing order-3 `test_compression` vectors still pass, confirming
  the rewrite is behaviour-preserving for previously supported orders.
- `cargo test -p rinex --lib hatanaka` passes.
- `cargo test -p rinex --lib` shows no new failures compared to `develop`
  (same pre-existing, unrelated `orbit_database_sanity` cascade failures
  in both cases).
- `rustfmt --check` on the changed file.

### DIFF
--- a/rinex/src/hatanaka/numdiff.rs
+++ b/rinex/src/hatanaka/numdiff.rs
@@ -48,7 +48,7 @@ impl<const M: usize> NumDiff<M> {
 
     /// Rotate internal buffer, take new sample into account.
     fn rotate_history(&mut self, data: i64) {
-        self.buf.copy_within(0..M - 2, 1);
+        self.buf.copy_within(0..M - 1, 1);
         self.buf[0] = data;
     }
 
@@ -152,5 +152,58 @@ mod test {
         assert_eq!(diff.compress(113483138205), -3150848442);
         assert_eq!(diff.compress(113469906136), 1575425367);
         assert_eq!(diff.compress(113457280090), 146);
+    }
+
+    #[test]
+    fn test_history_rotation_full_order() {
+        // Regression test for https://github.com/nav-solutions/rinex/issues/426
+        //
+        // `rotate_history` used to only shift `buf[0..M-2]`, so `buf[M-1]`
+        // (the oldest history slot) was frozen at its initial value forever
+        // and never took part in the running history. Decompression that
+        // reaches the maximal order `M` (as happens for the clock-offset
+        // field, which is decompressed with `NumDiff<5>` at order 5) then
+        // silently used a stale value instead of the correct one, causing
+        // decompression to diverge and eventually overflow further down
+        // the file.
+        //
+        // `compressed` below is the correct order-5 CRINEX encoding of
+        // `original`, computed independently of this crate. Decompressing
+        // it must reproduce `original` exactly, which requires every slot
+        // of the history buffer (including buf[M-1]) to stay up to date.
+        let original = [
+            126298057858_i64,
+            126282454570,
+            126267372371,
+            126252810509,
+            127814188268,
+            127800656941,
+            127787641437,
+            127775141621,
+            127762669477,
+        ];
+        let compressed = [
+            -15603288_i64,
+            521089,
+            -752,
+            1575420036,
+            -6301688027,
+            9452541607,
+            -6301698660,
+            1574937163,
+        ];
+
+        let mut diff = NumDiff::<5>::new(original[0], 5);
+
+        let mut recovered = vec![original[0]];
+        for value in compressed {
+            recovered.push(diff.decompress(value));
+        }
+
+        assert_eq!(
+            recovered, original,
+            "decompressed sequence diverged from the original: \
+             history rotation must keep every buffer slot (including buf[M-1]) up to date"
+        );
     }
 }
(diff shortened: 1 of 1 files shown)


gwbres:
@MrCry0, thank you very much for your thorough and detailed work, and adding a new test! 🚀

gwbres:
argh I did not catch that you pointed to the wrong branch, you should in the master branch

MrCry0:
> argh I did not catch that you pointed to the wrong branch, you should in the master branch

Ah, thanks! I missed this important point. :)

Well, a lot of this work already seems to be in master. My bad. I'll review my changes and prepare only the ones that are still missing from master.

Issue #413, opened 2026-04-18
Observation RINEX V4.02


gwbres:
as raised by #412, the RINEX V4.>01 revision introduces the possibility to increase the timestamping resolution to 1e-12, which can be particularly interesting for advanced navigation processes.

- [ ] obtain a reference file to test the process
- [ ] parse the timestamp correctly
  - [ ] [Hifitime is currently limited to nanoseconds](https://github.com/nyx-space/hifitime)

gwbres:
ping: @jmfriedt interesting asset.
ping: @ChristopherRabotin  internal limitations:
  - this can be a long term thing
  - this has been discussed a couple of times but not really tried either because we are focused on more important and urgent tasks. Last time we did, we observed - as expected - a decrease in performance. For me it's normal to observe such a decrease, for example, increasing to 1e-12 or 1e-21 as planned, will come with roughly a x8 or x4 decrease. But those applications are dedicated to post-processing and high performance computing, at some point, it needs to be decided what "performance" means and what the end goal is. As a reminder, 1e-9 limitation means limitations to 1m in a navigation process
  - recently I have been thinking about making the Epoch struct "generic". The epoch struct is quite simple (as it should) and that should be doable. We could have two implementations, one Ultra fast and less accurate, and one Slow and ultra precise, for example one with the current implementation, and the other with 128 bits. This way, you could have the best of both worlds. I'd like to see how other libraries deal with that in Rust. The only problem (for Hifitime Duration) is the scaling to 1 second, so we need someway to hardcode what is the actual Tick (time axis quanta) duration value, so we can express values correctly (the tick being 1e-9 currently)

ChristopherRabotin:
As you say, an epoch which is generic over two kinds of duration structures might be the solution: the Epoch used in 99% of the cases would use the current Duration, and for your specific use-case, you could use an `UltrapreciseDuration` which stores zeptoseconds as a single i128. In supporting both approaches, we could limit the initialization of this UltrapreciseDuration to strictly integers to bypass the issue of building the zepto duration from f64.

PR #431, opened 2026-09-03
hatanaka: fix receiver clock offset decompression (series 1/8)

### LINKED ISSUE #426: i64 overflow while decoding a valid CRINEX V3 receiver clock differential

MrCry0:
## Series: PR 1 of 8

This PR opens a series of eight stacked PRs. Together they make RINEX
observation, navigation, CRINEX and IONEX files parse and format
correctly across versions, add RINEX 4.01 and 4.02 support, make the
workspace build and test from a clean checkout, and make the `rinex-qc`
test suite deterministic. Each PR is based on the
previous one, so GitHub retargets it to `develop` once its base merges.
Every commit of the series builds and was tested with the whole
workspace; the failure count never rises from one commit to the next,
and it reaches zero at the tip of PR 4. Branches live on
`MrCry0/rinex`.

1. **PR 1, this one** (`fix/crinex-clock-offset`): decode the receiver
   clock offset of CRINEX files and fail with an error instead of a
   panic in `NumDiff`. Closes #426.
2. **PR 2, #432** (`pr2/workspace-builds-from-clean-checkout`): dependencies
   from crates.io, `rinex-qc` ported to gnss-rtk 0.8, loader fixes and
   README examples, so the workspace builds and tests without sibling
   checkouts.
3. **PR 3, #433** (`pr3/observation-and-crinex-correctness`): RINEX 2/3/4
   observation parsing and formatting fixes, CRINEX 1/3 decompression
   fixes and a working compressor, verified against RNX2CRX/CRX2RNX.
4. **PR 4, #434** (`pr4/observation-validity-and-doc-examples`): signal and
   epoch validity helpers fixed, every documentation example runs as a
   doctest.
5. **PR 5, #435** (`pr5/robust-almanac-setup`): almanac setup serialized
   across processes, optional JPL Earth orientation model, embedded
   almanac as offline fallback.
6. **PR 6, #436** (`pr6/ionex-parse-and-format-round-trip`): IONEX files parse
   completely (last map, RMS maps) and format back to the same record.
7. **PR 7, #437** (`pr7/rinex-4.0x-support`): RINEX 4.01 and 4.02
   files accepted, with the 4.02 navigation messages, subtypes and
   observation epoch extension parsed and tested on the specification
   examples.
8. **PR 8, #438** (`pr8/navigation-formatter`): navigation files
   written completely in RINEX 2, 3 and 4, converted between
   revisions, and every orbit field kept by the parser.

## Summary

Closes #426. Follow-up to #430, which fixed the `NumDiff` history
rotation but is not enough on its own: on `develop` the receiver clock
offset of a CRINEX file is never decoded at all, and on the released
code path it is decoded with an uninitialized kernel, which is what
overflowed in the report.

Four commits, each buildable and tested on its own (the last one only
lengthens the test extract).

### observation: store parsed receiver clock offsets

`parse_epoch()` extracted the clock offset from the epoch line, then
called `with_clock_observation()` (which returns a modified copy) and
dropped the result. No RINEX observation file ended up with any clock
offset in its record, compressed or not. The RINEX 2 branch was also
unreachable (`major < 2`) and split at the wrong column. The reused
`Observations` buffer now also clears its clock between epochs.

### hatanaka: fix receiver clock offset decompression

- `run_clock()` received the clock line with its `\n` still attached
  when driven by the file parser, so a kernel reset such as `3&1520`
  failed to parse and was silently ignored. The remaining branches only
  accepted single-character lines, so compressed values were dropped
  too (or, on a kernel that was never reset, integrated at the default
  order from an empty history: the divergence behind the overflow).
  The line is now trimmed and interpreted by content only: empty,
  `m&value` reset, or compressed value.
- The recovered integer was scaled as if it carried 3 decimals. CRINEX
  stores the clock offset with the decimal point removed: 1e-12 s for
  RINEX 3 (F15.12), 1e-9 s for RINEX 2 (F12.9). Verified against
  CRX2RNX 4.1.0, which prints `3&1520` as `.000000001520`.
- RINEX 2 output is padded to column 68 before the offset so it lands
  in columns 69-80 whatever the SV count; the V1 clock state size
  accounts for it.

### hatanaka: fail with an error instead of panicking in NumDiff

As suggested in the issue: `NumDiff::decompress()` / `compress()` now
return `Result<i64, Error>` and use checked arithmetic. An i64 overflow
yields `Error::NumericOverflow`, an order the kernel cannot serve
(above `M`, or above 6) yields `Error::CompressionOrder` instead of
indexing out of bounds; the `level > 6` panics in `new()` /
`force_init()` are gone. The decompressor propagates these, so the file
parser ends with what it recovered so far, as it does for any other
CRINEX error. The six binomial arms are replaced by a coefficient table.

## Test plan

New test resources (in-repo, `test_resources/`):

- `CRNX/V3/ACRG00GHA_R_20240010000_01H_30S_MO.crx` + `OBS/V3/...rnx`:
  one hour (120 epochs) of the BKG file from #426 reduced to a single
  SV (I02), re-encoded with RNX2CRX 4.1.0; the model is the CRX2RNX
  4.1.0 output. Its compressed clock lines are identical to the
  original file's. Covers the `3&` reset, the order-3 ramp-up, epochs
  without clock, and five further resets. One hour is the minimum that
  reproduces the report: on the released kernel this clock sequence
  overflows at epoch 91.
- `CRNX/V1/delf0010_clock.21d` + `OBS/V2/delf0010_clock.21o`: a RINEX 2
  extract of delf0010.21o with synthetic clock offsets (drift, a gap,
  a jump), same tooling. Exercises 2-character compressed values.

New tests (`rinex/src/tests/crinex/decompression/mod.rs`):

- `v3_acrg00gha_clock_offsets`: strict point by point comparison of
  every signal and clock offset against the CRX2RNX model
  (`generic_observation_comparison`), every clock offset compared
  again by epoch, plus hand-picked values around resets and the gap,
  and `generic_observation_rinex_test` with `has_clock = true`.
- `v1_delf0010_clock_offsets`: same clock checks on the RINEX 2 extract.
  Signals are not compared there: V1 observation decompression is
  already failing on `develop` (`testbench_v1` and friends), which is
  out of scope here.
- `v3_clock_overflow_does_not_panic`: the first reset of the extract is
  replaced by `3&9223372036854775807`; parsing returns normally with at
  most the first epoch instead of panicking.
- Both clock tests fail on unpatched `develop` ("model has no clock
  offsets"), fail with only the parser commit ("wrong number of clock
  offsets: 0"), and pass with the fix. Parsing the same one-hour
  extract on unpatched `main` (v0.23) panics with the exact message
  from the report, `attempt to multiply with overflow` in
  `NumDiff::decompress`, so the test covers both ways the issue shows
  up.

`rinex/src/hatanaka/numdiff.rs`: `test_overflow_is_an_error`,
`test_unsupported_order_is_an_error`, existing tests adapted to the
`Result` API. `test_parse_v3_1` now expects the zero-valued clock its
epoch line carries; `epoch_size_to_produce_v1` accounts for the clock
room.

Other checks:

- The full 24 h BKG file (5.3 MB, SHA-256 as in the issue) parses in a
  debug build without panic: 2880 epochs, 1584 clock offsets, all equal
  to the CRX2RNX 4.1.0 output.
- `cargo test -p rinex --lib`: 177 passed, and the same 31 pre-existing
  failures as `develop` (nav orbit database cascade, V1 observation
  decompression).
- `rustfmt --check` on every touched file.


gwbres:
@MrCry0 ❤️ ❤️ thank you for your the work, I will look into your previous post in details;

please drop a message here when you think you are ready to merge this 👍 

Once you're done I will release a new version of the RINEX lib, and related applications

MrCry0:
> @MrCry0 ❤️ ❤️ thank you for your the work, I will look into your previous post in details;
> 
> please drop a message here when you think you are ready to merge this 👍
> 
> Once you're done I will release a new version of the RINEX lib, and related applications

We may need to postpone this PR until I finish the full pack of fixes.

gwbres:
Hello @MrCry0,

the RINEX repo is now limited to the RINEX parser. It has been like this fore more than a year. 
Its "develop" branch is stall and does not stand any longer. It should be removed. I split the previous workspace into individual repos, which makes the maintenance a little more complex in some cases, but facilitates its management with "cargo". Indeed, "cargo" led to many troubles and issues in the past, due to its behavior in a workspace, especially in a workspace where we want libraries and applications, with intricate dependencies, and complex dependencies as demanded by GNSS navigation (nalgebra, nyx..etc..). Your work still applies and might be still relevant in most case. 

First off, I'm trying to understand whether all the "RINEX" (lib) topics have been already fixed or not (previous commits) I am looking into this in detail right now (PR1, PR3, PR4, PR7)

Can you check whether PR6 is still relevant and would apply to [the IONEX library?](https://github.com/nav-solutions/ionex)

PR2 will not apply and is outdated, it simply does not stand

PR5 (`gnss-qc` and related applications like rinex-cli) is the most problematic. The current framework state, is basically I fixed the cargo workspace and created individual components, and even improved (and had great contributions) the core libraries, like RINEX or IONEX. But `gnss-qc` is kind of old and not up to date. This library is a intermediate level glue, to allow build an application like rinex-cli, basically by allowing combining several input files. Finally, rinex-cli is completely broken, because it must be updated to the latest stuff, in particular, I made great progress in GNSS-RTK, the navigation core, but the interface has changed dramatically, this step has not been taken yet

MrCry0:
Thanks for the update. I checked each PR against the current main by running its code (data submodule included) and kept only what is still broken there. PR 2 (#432) and PR 5 (#435) are closed. PR 4 is mostly moot on main (LLI check correct, doctests pass); PR 6 does not apply to nav-solutions/ionex, which parses every map and round-trips records (only the header station/satellite counts are lost on write, I can send a one-line PR there).

Resent against main, small and independent where possible:
- #439 header: fractional TIME OF FIRST/LAST OBS parsed wrong (found while bumping the data submodule, the new 2024_09_20 file trips it); bumps the submodule to data main
- #440 observation: clock offset formatter (F15.12 / F12.9)
- #441 hatanaka: NumDiff rotation and compress fix (#429) plus a Result API instead of panics
- #442 navigation: writer fixes (orbit line indent, SBAS week, RINEX 4 STO/EOP/ION bodies, RINEX 2 layout)
- nav-solutions/data#6: test extracts for #426

The #426 fix itself (clock offsets decoded, stored and compressed) is stacked on #440 and #441 and needs the data PR; I will open it once these are merged.

Still to come: RINEX 4.02 navigation (main parses 15 of the 30 spec records), the 4.02 picosecond field, and the CRINEX V1 observation line issue.

MrCry0:
wrong base branch

gwbres:
@MrCry0, please don't delete your work, we just need to take each PR one by one and see how it would apply to the main branch

I saw you provided test Clock Data (which is awesome, the lib does not have it yet!). In the current setup, I regrouped all test data [under this repo](https://github.com/nav-solutions/data/tree/main) which is used as submodule everywhere

MrCry0:
Follow-up now that #439, #440, #441, #442 and #445 are merged (thanks). The remaining pieces, rebased on the current main:

- #443 navigation: RINEX 4.02 support (message types, subtypes, NavIC and GLONASS CDMA ephemerides and ionosphere models, blank PRN headers, STO time system pairs, orbit field alignment, 4.01 observation codes). Carries the data submodule bump to the two merged data PRs.
- #446 navigation: writing a record in another revision than the parsed one (RINEX 4 to 3 keeps the legacy ephemerides instead of aborting on the first CNAV record, RINEX 2/3 to 4 gets the message types). Stacked on #443, merge after it.
- #447 the #426 fix itself: receiver clock offsets decoded, stored and compressed. Stacked on #446, merge after it.
- #444 hatanaka: CRINEX 1 decompression (independent).
- nav-solutions/ionex#12: the answer to your PR 6 question. Only the header station and satellite counts were lost on write; the PR adds them and fixes the two unit tests and three doctests that fail on ionex main.

Still to come after #447: the RINEX 4.02 picosecond epoch field of observation files.

gwbres:
I updated the base branch so it is ready to be released any time (core dependencies).

Personnally I have no opinion on when to release, if you feel like continue pushing more patches, we can wait and have all of this integrated in the next version. If you need or want an intermediate version, that can also happen. To propose a new release, all you need to do is open a PR with the Cargo version update, and (once merged & verified) you simply push a tag, this will trigger the cargo publish action. I have done it myself all the time, but any other contributor can do that actually ;)

> Still to come after https://github.com/nav-solutions/rinex/pull/447: the RINEX 4.02 picosecond epoch field of observation files.

Someone mentioned V4.02 a few months ago, and I summarized it in this issue: #413. 

My opinion is, it's great if the parser is updated, that means the framework is ready to process it correctly. The problem you will face is that the core library (hifitime) is limited to 1ns, so even though you manage to catch the new timestamp format correctly, you will not be able to process it correctly

MrCry0:
Agreed, and it shows up concretely: hifitime 4.3.1 (pinned across the whole workspace, still the latest release on crates.io) stores Duration as centuries plus a u64 nanosecond count, so 1ns is the real floor no matter what the parser accepts.

What I have locally treats the picosecond field as syntax, not precision: the extra "1X,I5.5" digits get parsed so the reader does not choke on a 4.02 observation file, but everything past 100ns is floored on read, and comes back as zeros on write (nanos % 100 * 1000). So it reads and re-emits the field without crashing or losing the seconds-to-100ns part, but it cannot carry or use the picosecond information for anything. Same limitation you're describing.

Given that, I don't think it makes sense to wait on hifitime for this. My plan is to scope it to what's honestly achievable now: store the raw picosecond digits as their own field on the epoch, separate from the Duration/Epoch used everywhere else, so a file that carries them round-trips byte for byte. No claim of picosecond-accurate processing anywhere in the crate, just no silent data loss on parse to format. If hifitime (or something else) eventually gets sub-ns Duration, that field is the input to wire up properly; until then it stays inert everywhere except the observation epoch line.

Let me know if you'd rather this stay out of the parser entirely until hifitime moves - happy to hold it as its own PR either way.

gwbres:
> What I have locally treats the picosecond field as syntax, not precision: the extra "1X,I5.5" digits get parsed so the reader does not choke on a 4.02 observation file, but everything past 100ns is floored on read, and comes back as zeros on write (nanos % 100 * 1000). So it reads and re-emits the field without crashing or losing the seconds-to-100ns part, but it cannot carry or use the picosecond information for anything. Same limitation you're describing.

Yes that's what I meant, great

> Given that, I don't think it makes sense to wait on hifitime for this. My plan is to scope it to what's honestly achievable now: store the raw picosecond digits as their own field on the epoch, separate from the Duration/Epoch used everywhere else, so a file that carries them round-trips byte for byte. No claim of picosecond-accurate processing anywhere in the crate, just no silent data loss on parse to format. If hifitime (or something else) eventually gets sub-ns Duration, that field is the input to wire up properly; until then it stays inert everywhere except the observation epoch line.

If you have a proposal to go as far as garantee bytewise correctness, maybe both ways since you seem to be interested in RINEX formatting, I am totally cool with that. Currently I don't foresee myself getting rid of Hifitime because that would mean getting rid of other astro/navigation features that 100% rely on this core library (mainly, orbit determination). But in a mid/long term, that may become an issue. The current discussion in Hifitime is that we could propose two implementations: preserve the current, and propose a "ultra precise" (at the expense of more computation load) implementation. The idea is to preserve the current performance. That seems doable, but it will be a long term thing

PR #432, opened 2026-09-07
workspace: build and test from a clean checkout (series 2/8)

MrCry0:
## Series: PR 2 of 8

Second PR of the series opened by #431 (see its description for the
whole list). Stacked on #431: the four commits of PR 1 show in the
diff until it merges. PR 2 touches none of the files PR 1 touches.

## Summary

On `develop` the workspace manifest cannot be loaded on a machine
without the maintainer's sibling checkouts (`../../rtk-rs/gnss-rtk`,
`../../cggtts/cggtts`), so nothing builds, not even
`cargo build -p rinex`. This PR makes the workspace build and test
from a clean checkout. Six commits:

### navigation: name the SBAS orbit database entries SBAS instead of GEO

`db/NAV/orbits.json` named the two SBAS frames "GEO". The build script
turns that into `Constellation::from_str("GEO").unwrap()`, which fails
with gnss-rs 2.6, so the first access to the database panicked and
poisoned the lazy static behind every navigation test. The parser
already normalizes SBAS vehicles to `Constellation::SBAS`, so the
entries only had to be renamed.

### rinex-qc: port to gnss-rtk 0.8 and take dependencies from crates.io

gnss-rtk 0.8.0 and cggtts 4.4.0 from crates.io, gnss-rs from crates.io
everywhere, rinex/sp3/rinex-qc moved to anise 0.6 (gnss-rtk 0.8
depends on it; two anise copies made the solver types incompatible).
The `rinex-qc` navigation solvers are ported to the gnss-rtk 0.8
generic `Solver` (orbit, bias, ephemeris and time sources implemented
on top of the ephemeris context). `SignalSource::collect_epoch`
stamped signals with the epoch of the following one; the ported
solver only resolves within range of the station once that is fixed,
so the fix is part of the same commit. The four pieces form one
commit because `rinex-qc` does not compile between them.

### rinex-qc: route gzip files through the gzip loader in load_file

`load_file` is documented as detecting the format but sent `.gz`
paths to the plain readers.

### rinex-qc: take the full suffix as the file extension in MetaData

Compressed SP3 files were indexed with extension "gz" while compressed
RINEX got "crx.gz"; the name is now split at its first dot.

### rinex-qc: fix the README examples

The README is the crate documentation and its examples run as
doctests; they called APIs that do not exist. Rewritten against the
current API, with the sections about features the solver does not
provide yet saying so.

### rinex: add the observation file of the dominant sample rate test

`240506_glacier_station.obs.gz` was never carried over to this
branch's test resources although the test that reads it was. Restored
from the v0.17 history (byte-identical blob, 4.75 MB).

## Test plan

- Every commit was built and tested with
  `cargo test --workspace --all-features` from a clean worktree. The
  first commit cannot load the workspace, like `develop`; from the
  port commit on, the whole workspace compiles and the failure count
  never rises.
- At the tip: 340 tests pass, 39 fail. The 39 are pre-existing and
  owned by the next PRs: 8 rinex unit tests (CRINEX V1 and RINEX 2
  observation, PR 3), 30 rinex doctests and 1 sp3 doctest (PR 4).
  `rinex-qc` is fully green: 22 unit tests and 7 README doctests.
- The 20 configurations of the CI build matrix build at the series
  tip; `cargo fmt --all -- --check` passes at every commit;
  `tools/build-docrs.sh` builds the documentation without new
  intra-doc link warnings.
- `rinex-cli` builds and checks with default and all features once
  `rinex-qc` compiles; it stays a workspace member.


gwbres:
Hello, the workspace does not stand any more, this branch is completely stall do not work on this.
The workspace was split into each individual repo that exists in this framework online.
The cargo workspace was a pain to deal with and led to many troubles.
Only the master branch and maybe a few, still hold here.
The develop branch can be erased and was kept just to make sure previous content was not lost in the duplication process

MrCry0:
Closing: the workspace and the develop branch no longer stand (see the comment above). The remaining fixes of the series will be resent against main.

PR #433, opened 2026-09-07
observation: RINEX 2/3/4 and CRINEX 1/3 correctness (series 3/8)

MrCry0:
## Series: PR 3 of 8

Third PR of the series opened by #431 (see its description for the
whole list). Stacked on #432: the ten commits of PR 1 and PR 2 show in
the diff until they merge.

## Summary

Parsing and formatting of observation RINEX 2, 3 and 4 and of CRINEX 1
and 3, verified against RNX2CRX/CRX2RNX 4.1.0. Eight commits, `rinex`
crate only:

### hatanaka: fix CRINEX V1 decompression

Observation lines are decoded as exactly `numobs` blank-separated
fields followed by the LLI/SNR flag text, as CRINEX defines them,
instead of ad hoc rules on the next blank. The V1 epoch description
emitted one continuation line too many from 11 SVs on, and the record
parser stacked compressed lines that produce no output as raw text,
which the positional RINEX 2 parser consumed as observations. Every V1
file lost its alignment right after the epoch line.

### observation: fix RINEX 3 signal field parsing

Fields were sliced from column 5 instead of 4, so a value with an LLI
flag got the flag digit as a fourth decimal. The SNR digit was mapped
as a dB-Hz value instead of a RINEX code.

### formatting: write standard header position, system letter and epochs

`APPROX POSITION XYZ` in 3F14.4 instead of 14 decimals overflowing the
field, the satellite system letter in column 41 of the version line,
zero padded epoch seconds.

### hatanaka: reset returning vehicles and drop stale epoch descriptors

A vehicle absent from the previous epoch restarts from scratch in
CRINEX; the decompressor kept its old text kernel. The epoch kernel
never shrinks, so identifiers of an earlier, larger epoch trailed the
description and the V1 formatter printed them as observation data.

### hatanaka: implement the CRINEX compressor

The compressor wrote a placeholder for every observation. It now
compresses the way RNX2CRX does (text kernel for the epoch and flags,
order 3 numerical kernels reset with `m&value`, trimmed lines).
`TextDiff::compress` mirrors the decompressor.

### observation: keep the leading blank of RINEX 2 SV lists

Files that omit the constellation letter lost their leading blank and
every SV slot shifted, dropping the last vehicle of every epoch.

### observation: format SBAS vehicles with the SBAS observables

The formatters looked observables up with the vehicle's own system
(EGNOS, WAAS, ...) while the header keys them as SBAS, and failed on
any file carrying an SBAS vehicle.

### tests: cover RINEX 4 observation decompression and round trip

A one hour ACRG00GHA extract with a 4.00 version line, compressed with
RNX2CRX 4.1.0, decompresses to the CRX2RNX output and round trips.

## Test plan

  --all-features` from a clean worktree; the failure count never
  rises.
- Unit failures go from 8 at the PR 2 tip to 0 here: the four CRINEX
  V1 tests, `obs_header_formatting`, `merge_obs`, `v2_kosg0010_95o`
  and `v2_aopr0010_17o`.
- Round trips: the RINEX 2 and 3 model files compressed by this crate
  parse back to the same signals and clock offsets, and the same
  CRINEX files decompressed with CRX2RNX 4.1.0 yield the model data
  (zegv0010, npaz3550, delf0010_clock, ACRG00GHA, ACOR00ESP,
  pdel0010).
- Test expectations corrected against the model files are listed in
  each commit body (testbench_v1, test_parse_v3_1, KOSG0010.95O,
  merge_obs).
- At the tip: 353 tests pass, 31 fail, all documentation examples
  fixed by PR 4 (30 rinex doctests, 1 sp3 doctest).


MrCry0:
wrong base branch

PR #434, opened 2026-09-07
rinex: observation validity helpers and documentation examples (series 4/8)

MrCry0:
## Series: PR 4 of 8

Fourth PR of the series opened by #431 (see its description for the
whole list). Stacked on #433: the 18 commits of PR 1 to PR 3 show in
the diff until they merge.

## Summary

Observation validity helpers that rejected everything, and every
documentation example of `rinex` and `sp3` running as a doctest. Eight
commits:

### rinex: return an error when a header line cannot be read

The header parser unwrapped every line read, so a file that is not
valid UTF-8 (a gzip stream opened with `from_file`) panicked. A new
`ParsingError::IO` variant propagates the read error.

### rinex: report epochs with abnormal flags in epoch_anomalies

The filter kept the epochs whose flag is OK, the opposite of what it
documents.

### rinex: add LliFlags::is_ok and use it for phase tracking checks

`LliFlags::OK_OR_UNKNOWN` is the empty bit set, so `intersects()`
against it is always false: `phase_range_tracking_ok_iter` dropped
every flagged phase observation and `is_ok_snr` accepted none.
`LliFlags::is_ok` is true when neither a lock loss nor a half cycle
slip is reported.

### rinex: use a XXXX placeholder in short names of unknown stations

Short standard names emitted "0070.21M" for an unknown station; the
documented placeholder is "XXXX".

### rinex: fix the documentation examples

Most README and API examples did not compile (renamed methods,
changed iterator items, header fields that became `Option`, missing
files, "example" and "TODO" placeholders, a merge conflict marker).
Rewritten against the current API and the test resources.

### sp3: fix the interpolation example of the README

The example tested an `Option` with `is_err` and interpolated at the
first epoch; it now uses a feasible epoch of the file.

### rinex: accept signals without SNR in SignalObservation::is_ok

A missing SNR was replaced by the default value, which is never
strong, so every observation without SNR was rejected, against the
documented behaviour.

### rinex: check the phase tracking filter on flagged observations

The tracking filter test gains a file carrying lock loss and half
cycle slip flags and guards against a data set without sane flags.

## Test plan

- Every commit built and tested with `cargo test --workspace
  --all-features` from a clean worktree; the failure count never
  rises.
- Failures go from 31 at the PR 3 tip to 0 at this tip: 389 tests
  pass, including 53 `rinex` doctests, 3 `sp3` doctests and 7
  `rinex-qc` doctests. This is the first tip of the series where the
  whole workspace is green.
[...]

PR #435, opened 2026-09-07
rinex-qc: robust almanac setup (series 5/8)

MrCry0:
## Series: PR 5 of 8

Fifth PR of the series opened by #431 (see its description for the
whole list). Stacked on #434: the 26 commits of PR 1 to PR 4 show in
the diff until they merge.

## Summary

Deterministic almanac setup for `QcContext`, so the `rinex-qc` suite
passes on a cold checkout and without network. Five commits,
`rinex-qc` only:

### rinex-qc: serialize the almanac setup across processes

Every context downloaded and stored the almanac on its own, and tests
run in parallel processes and threads. anise guards the download with
a lock file that is checked then created, which does not exclude
concurrent writers, and the storage description was written in place.
An exclusive file lock (`std::fs::File::lock`, stable since Rust 1.89,
so the crate `rust-version` moves from 1.64 to 1.89) makes processes
take turns; the loaded almanac is shared within a process; the
description is written to a temporary file and renamed.

### rinex-qc: make the daily JPL Earth orientation model optional

On a cold checkout each process hit the 30 s download timeout of the
JPL file in turn and every test failed. The two nyxspace files stay
mandatory, the JPL model is fetched when the storage is created and
skipped with a warning otherwise. The ITRF93 frame is only selected
when that model loaded: without it the solver cannot transform any
satellite state, while the IAU Earth frame works.

### rinex-qc: fall back to the embedded almanac when the download fails

The documented offline model: `Almanac::until_2035` with the IAU
Earth frame when the mandatory files cannot be retrieved, nothing
stored so the download is retried next time.

### rinex-qc: remove leftovers of interrupted anise downloads

anise leaves an empty file and its lock behind when a download times
out while reading the body; the next attempt waits ten seconds on
that lock. Both are removed when the lock is older than the anise
download timeout. The download location is computed with
`platform-dirs`, the crate anise uses for it (MIT, already in the
dependency tree).

### rinex-qc: retry the JPL model once per local storage

A storage written during a JPL outage stayed on the lower precision
frame until removed by hand; the model is now retried once per
storage, recorded in a marker file.

## Test plan

  --all-features` from a clean worktree; all green at every commit
  (392 tests at the tip).
- Cold scenarios on a fresh storage: with network, all `rinex-qc`
  tests and doctests pass in parallel while JPL timeouts are absorbed;
  with all downloads blocked, the embedded almanac takes over and the
  suite passes in about 13 s.
- New tests: `concurrent_deployment` (8 contexts at once),
  `stale_download_removal` (four leftover scenarios),
  `jpl_model_retry` (retry policy); the PVT solutions test passes on
  the IAU frame within its 100 m bound.
- Behaviour changes: a missing network no longer fails context
  deployment; ITRF93 is reported only when the JPL model is present;
  `rinex-qc/.cache/` is ignored as a whole.


MrCry0:
Closing: rinex-qc is no longer part of this repository (see the discussion in #431). The almanac setup changes would have to be redone against nav-solutions/gnss-qc once it is brought up to date.

PR #436, opened 2026-09-07
ionex: parse every map and format files that parse back (series 6/8)

MrCry0:
## Series: PR 6 of 8

Sixth PR of the series opened by #431 (see its description for the
whole list). Stacked on #435: the 31 commits of PR 1 to PR 5 show in
the diff until they merge.

## Summary

IONEX files parse completely and format back to the same record.
Three commits, `rinex` crate only:

### ionex: parse the last map of a file and the RMS maps

The record parser decided how to parse a buffered IONEX block from
the START line of the next block instead of the block's own, so the
last map of every file was never parsed and, when RMS maps follow the
TEC maps, the last TEC map went to the RMS parser and the first RMS
map to nothing. The RMS parser also read its grid line from the whole
block instead of the current line and failed on every RMS map. Every
IONEX file lost its last map and never had RMS values.

### ionex: format a header the parser reads back

The header formatter wrote no version line for IONEX, "TODO" for both
map epochs, and omitted the number of maps, base radius and exponent.
Every field the parser reads is now written in the layout of the
specification. Two Display implementations that recursed forever now
write their labels, and the header parser stores `# OF SATELLITES`
for IONEX instead of swallowing it in an observation branch.

### ionex: write TEC and RMS maps the parser reads back

The record formatter never dispatched IONEX records, so a written
IONEX file was a header alone; the map formatter it would have called
numbered every map 0, wrote no epoch line, never entered its grid
loop and wrote floats instead of scaled integers. Maps are now
numbered from 1 with their epoch, the grid is browsed as the header
defines it, values are I5 integers in the header exponent, RMS maps
follow the TEC maps and the file ends with `END OF FILE`. Height maps
are not written: there is no data model and the parser skips them.

## Test plan

  --all-features` from a clean worktree; all green at every commit
  (396 tests at the tip).
- `v1_complete_maps`: all four IONEX test files parse the number of
  maps their header announces, with an RMS value per point for the
  JPL file. Fails on the previous commit (24 of 25 and 12 of 13 maps,
  no RMS at all).
- `ionex_header_formatting`: formatted header lines match the
  original files byte for byte and parse back to the same fields.
- Round trips: CKMG0020 (25 TEC maps) and jplg0010 (13 TEC plus 13
  RMS maps) format, parse back and compare equal as records, with map
  numbering and epochs checked on the text.
- The test time frames of the three existing IONEX tests were
  extended to the last map, and the RMS expectations of their test
  points are now asserted.
- Known gaps left in place: the mapping function is never parsed (the
  parser matches a misspelled marker), map dimension is not parsed,
  differential code biases are neither parsed nor written.


MrCry0:
wrong base branch

PR #437, opened 2026-09-07
rinex: support RINEX 4.01 and 4.02 (series 7/8)

MrCry0:
## Series: PR 7 of 8

Seventh PR of the series opened by #431 (see its description for the
whole list). Stacked on #436: the 34 commits of PR 1 to PR 6 show in
the diff until they merge.

## Motivation

`Version::is_supported` accepted minor revisions up to 4.00 only, so every 4.01 or 4.02 file (for example the current CDDIS rapid navigation products) was rejected with `NonSupportedVersion` before its header was even parsed. Beyond the version gate, RINEX 4.01/4.02 add a few format deltas that the parser did not know about.

## Changes

Version handling
- Any minor revision of a supported major revision is accepted, including minor revisions newer than this library. Unknown major revisions are still rejected. `Version::default()` and the revision written into new files stay 4.00.

Observation (4.02 Table A3)
- Optional picosecond extension of the epoch seconds, parsed (floored to the 1 ns `Epoch` resolution) and written back for 4.02+ files.

Navigation (4.02)
- Record header message subtype (WIDE, JAPN, KLOB, NEQN), stored in `NavKey` and written back.
- Message types L1NV, L1OC, L3OC, LXOC.
- NavIC LNAV and L1NV ephemeris definitions (Tables A30, A31); NavIC records were not parsed at all before (no orbit definition, no timescale).
- GLONASS L1OC and L3OC CDMA ephemeris definitions (Tables A16, A17).
- NavIC L1NV Klobuchar and NeQuick-N models and the GLONASS CDMA model (Tables A38 to A40) as new `IonosphereModel` variants.
- Optional CNAV/CNV2 integer flags field (Tables A10, A11, A20, A21).
- Record headers naming the constellation only (blank PRN), represented by PRN 0.

Observation codes (4.01)
- GPS `R` attribute (L1R, L2R) and NavIC L1 codes for carrier identification.

Pre-existing bugs found while testing against the specification examples, fixed here since the 4.02 tests depend on them
- The STO message line was read as A0, A1, A2, t_tm instead of t_tm, A0, A1, A2 (Table A33).
- The V3 observation formatter wrote the receiver clock offset with four decimals instead of F15.12.
- A blank orbit field followed by omitted trailing fields on the same line shifted the fields of the following lines.

## Tests

- Version: the V4 observation, CRINEX and navigation resources rewritten in memory as 4.01 and 4.02 must parse to the same record as 4.00.
- Every new navigation message is tested on the specification examples (Tables A12, A18, A32, A41), both as unit tests and through a new 4.02 navigation resource gathering all of them.
- Formatting round trips for the clock offset and the picosecond field.

## Notes for reviewers

- API changes: `NavKey` gains a `subtype` field and `IonosphereModel` gains three variants. Code constructing `NavKey` literals or matching `IonosphereModel` exhaustively needs an update.
- Known limitation, not addressed here: STO records sharing epoch, satellite and message type share a `NavKey`, so only the last one is kept (the Table A41 examples give 14 records for 9 keys).
- NavIC epochs are expressed in GPST; the NavIC week count runs from 1980 like GPS (Table A30).


MrCry0:
wrong base branch

PR #438, opened 2026-09-07
navigation: write complete records and convert between revisions (series 8/8)

MrCry0:
## Series: PR 8 of 8

Last PR of the series opened by #431 (see its description for the
whole list). Stacked on #437: the 49 commits
of PR 1 to PR 7 show in the diff until they merge.

## Summary

The navigation formatter did not write navigation files. Reported
while testing #437: every written ephemeris was reduced to its
satellite and epoch, all of them on one line, and the result could not
be read back. Five commits, `rinex` crate only.

### navigation: keep every orbit field present in the record

`parse_orbits` meant to drop orbit fields set to zero, but tested for
the substring `.000000000000E+00`, which every integer valued field
with a zero exponent also contains. Health flags, L2 codes, accuracy
indexes, fit intervals or issue of data equal to 1 to 9 were dropped
whenever the file uses an uppercase exponent, and kept with a lowercase
or `D` exponent. Every field present in the record is now stored;
`Ephemeris::get_orbit_f64` still hides zero values.

### navigation: keep health and status values the enums do not describe

GPS health 63 became 0, GLONASS health 5 became 4, and the GLONASS
status and Galileo health flags lost every bit they do not name.
Values outside the enums are kept as raw integers (`OrbitItem::U32`)
and undescribed bits are retained. Constellations with no health enum
(BeiDou) get the raw integer too, where the conversion used to be
declared unreachable.

### navigation: write complete records in every revision

Records are written as the parsers read them: E19.12 fields with a `D`
exponent in RINEX 2 and `E` from RINEX 3 on, orbit lines in the order
of the orbit database for the revision and message type, blank fields
for spare or absent items, the SBAS week in the clock drift rate slot,
and the STO, EOP and ION records of RINEX 4 (Tables A33 to A40,
including the 4.02 NavIC and GLONASS CDMA models). Records are written
in key order, which is chronological. The RINEX 2 epoch line uses the
two digit PRN and year, the RINEX 4 STO, EOP and ION epoch lines four
leading blanks.

### navigation: write records in a revision other than the parsed one

- Written as RINEX 2 or 3, only the ephemerides of the legacy messages
  are kept (LNAV, FDMA, INAV, FNAV, D1, D2, SBAS), laid out with the
  RINEX 3 definitions. CNAV, CNV1 to CNV3, L1NV, L1OC, L3OC and the
  STO, EOP and ION records have no representation there and are
  skipped, as are the satellites of other constellations in a single
  constellation RINEX 2 file.
- Written as RINEX 4, ephemerides parsed from RINEX 2 or 3 get their
  RINEX 4 message type: FDMA for GLONASS, SBAS for SBAS, INAV or FNAV
  for Galileo from the data source flags (Table A13), D1 or D2 for
  BeiDou from the GEO PRN range (BDS ICD, the specification gives no
  rule).
- The BeiDou group delays are named differently by the RINEX 3 and 4
  orbit definitions (`tgd1b1b3` / `tgdb1b3`) and are written under
  either name.

### navigation: fail when no frame can be written in the target revision

A record that is not empty but has no frame representable in the
target revision (only STO, EOP, ION or modern messages, or only other
constellations for a single constellation RINEX 2 file) returns
`FormattingError::NoRepresentableFrame` before anything is written,
instead of an empty record and success. One legacy ephemeris is enough
for the write to succeed, and an empty record still writes an empty
file.

## Test plan

  --all-features` from a clean worktree; all green at every commit.
- Round trips at the parsed revision: `cbw10010.21n.gz`, `amel0010.21g`,
  `dlf10010.21g`, `ijmu3650.21n.gz` (RINEX 2), `AMEL00NLD`, `CBW100NLD`,
  `BRDC00GOP`, `ESBC00DNK` (RINEX 3), `KMS300DNK`, `BRD400DLR` (RINEX 4,
  every message type) and the 4.02 examples file: the written file
  parses to the same header revision, constellation and record.
- Text checks: a RINEX 2 GPS and a GLONASS record are written byte
  for byte like the originals (`D` exponent, two digit PRN and
  year); the KMS STO record layout; every record of the RINEX 4.02
  examples file compared to the specification text (case and trailing
  blanks aside).
- Conversions: KMS (RINEX 4) written as 3.05 keeps the 357 legacy
  ephemerides, each field equal to the RINEX 4 value; AMEL (RINEX 3)
  written as 4.00 gets the expected message types with every field
  preserved, and written back as RINEX 3 gives the original record;
  CBW (RINEX 3, mixed) written as a 2.11 GPS file keeps the GPS
  ephemerides only.
- `NoRepresentableFrame`: the 4.02 examples reduced to their modern
  messages and STO/EOP/ION fail as 3.05 and succeed as 4.02; the
  GLONASS ephemerides alone fail as a 2.11 GPS file and succeed as a
  GLONASS one; an empty record succeeds.
- Parser regressions: a GPS LNAV record with uppercase exponent keeps
  its health, L2 codes, P data flag, accuracy and fit interval; GPS
  health 63 and GLONASS status 147 survive.
- `cargo fmt --all -- --check` passes at every commit; no new
  documentation or clippy warnings in the touched files.

## Notes for reviewers

- Known limitation: RINEX 3 files every ephemeris as LNAV, so a
  Galileo INAV and FNAV pair of the same epoch written as RINEX 3
  parses back to one key (the INAV one, written last). Same for the
  STO records sharing epoch, satellite and message type (#437).
- Behaviour change of the parser: zero valued orbit fields are now
  stored. A GPS vehicle with health 0 (every signal OK) thus gets
  `orbits.get("health") == Some(Health::Unhealthy)`, since the crate's
  `Health` enum names value 0 `Unhealthy`; nothing in the workspace
  consumes that item, and the enum naming is left for a follow-up.
  `get_orbit_f64` is unchanged.


MrCry0:
wrong base branch

PR #439, opened 2026-09-07
header: scale the TIME OF FIRST/LAST OBS fraction to nanoseconds

MrCry0:
## Summary

`parse_time_of_obs()` read the digits after the decimal point of `TIME OF FIRST OBS` / `TIME OF LAST OBS` as a plain integer and printed them over eight positions, so the seven digit fraction of the F13.7 field lost a factor of ten: `50.9890000` parsed as `50.0989`, and writing the header back shifted the digits once more. Two commits:

- `header: scale the TIME OF FIRST/LAST OBS fraction to nanoseconds`: scale the digits according to how many were given.
- `data: bump submodule to the current data main`: adopts the current content of nav-solutions/data (93250cd). Among the new files, `OBS/V3/2024_09_20_10_17_06.obs.gz` has a fractional time of first observation and fails the existing `production::obs_v3` formatting round trip without the first commit.

## Test plan

- `parse_time_of_obs` unit test extended with a seven digit fraction and a one digit fraction.
- `cargo fmt --all -- --check` and `cargo test --all-features` green at both commits (209 tests, 12 ignored), including `production::obs_v3` with the new data.

## Notes

This is the first of a series resent against `main` after the develop branch was retired (#431). The later PRs bump the submodule further; without this fix their `obs_v3` round trip fails on the new file.

### DIFF
--- a/src/header/parsing.rs
+++ b/src/header/parsing.rs
@@ -820,7 +820,7 @@ impl Header {
         let (mm, rem) = rem.split_at(6);
         let (ss, rem) = rem.split_at(5);
         let (_dot, rem) = rem.split_at(1);
-        let (ns, rem) = rem.split_at(8);
+        let (fraction, rem) = rem.split_at(8);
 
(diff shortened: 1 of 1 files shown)


gwbres (approved):
Thank you for all the comments and trying to keep this tool as clear as possible (in spite of the technical challenge 😆)🚀

PR #440, opened 2026-09-07
observation: format the receiver clock offset in the epoch line

MrCry0:
## Summary

The RINEX 3 observation formatter wrote the receiver clock offset with four decimals (`{:13.4}`), so offsets in the nanosecond range came out as `0.0000` and were lost when the file was read back. The RINEX 2 formatter did not write it at all (`TODO format clock data`).

Write the field as F15.12 in columns 42-56 for RINEX 3, and as F12.9 in columns 69-80 of the first epoch line for RINEX 2, after the twelve SV slots padded with blanks. One commit.

## Test plan

- Unit tests on `format_epoch_v2` (three vehicles, and fourteen vehicles with the offset on the first line) and `format_epoch_v3` (positive and negative offset).
- `cargo test --all-features` green (211 tests).

## Notes

No file in the repository exercises the change end to end yet: the parser drops every clock offset today. The clock offset PR (stacked on this one and on the NumDiff PR) stores them, at which point the existing `production::obs_v3` round trip covers this formatter on the VLNS files.

### DIFF
--- a/src/observation/formatting.rs
+++ b/src/observation/formatting.rs
@@ -126,39 +126,49 @@ impl Observations {
         const NUM_SV_PER_LINE: usize = 12;
         const NEW_LINE_PADDING: &str = "                                ";
 
-        if let Some(clock) = self.clock {
-            // TODO format clock data!
-            write!(
-                w,
-                " {}  {} {:2}",
-                epoch_format(key.epoch, RinexType::ObservationData, 2),
-                key.flag,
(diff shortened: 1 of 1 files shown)


gwbres (approved):
@MrCry0 awesome! this topic has been on standby since day-1, mostly due to lack of data ! 🚀 

Thank you for taking the time to write tests it really helps 👍

PR #441, opened 2026-09-07
hatanaka: fix NumDiff history rotation, return errors instead of panicking

MrCry0:
## Summary

Same changes as #430 (merged on the retired develop branch) plus the error handling that followed it. Two commits:

- `hatanaka: fix NumDiff history rotation and compress symmetry`: `rotate_history()` shifted only `buf[0..M-2]`, so the oldest slot never took part in the history and any (de)compression reaching order M diverged until an arithmetic overflow panic (#426 on CRINEX V3 files with clock offsets). `compress()` also rotated before differencing and indexed `buf[M]` at the maximal order (#429).
- `hatanaka: fail with an error instead of panicking in NumDiff`: `decompress()` and `compress()` return `Result`; checked arithmetic yields `Error::NumericOverflow`, an order beyond M or 6 yields `Error::CompressionOrder` instead of an out of bounds index. The decompressor propagates the error, the compressor gets a `FormattingError::Hatanaka` variant.

## Test plan

- Regression tests with an order 5 sequence computed independently of the crate (decompression, compression, round trip), overflow and unsupported order tests.
- `cargo test --all-features` green at both commits (212 and 214 tests).

## Notes

API change: `NumDiff::decompress` and `NumDiff::compress` return `Result<i64, hatanaka::Error>`; `hatanaka::Error` is now `Clone + Copy + PartialEq`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -201,6 +201,9 @@ pub enum FormattingError {
 
     #[error("nav: missing grid defs")]
     NoGridDefinition,
+
+    #[error("crinex compression: {0}")]
+    Hatanaka(#[from] HatanakaError),
 }
 
 /// General error (processing, analysis..)
--- a/src/hatanaka/compressor.rs
+++ b/src/hatanaka/compressor.rs
@@ -186,7 +186,7 @@ impl<const M: usize> CompressorExpert<M> {
                             .filter(|((sv, obs), _)| *sv == signal.sv && obs == &signal.observable)
                             .reduce(|k, _| k)
                         {
-                            let compressed = sv_kernel.compress(quantized);
+                            let compressed = sv_kernel.compress(quantized)?;
                             write!(w, "{} ", compressed)?;
                         } else {
                             // first encounter: build kernel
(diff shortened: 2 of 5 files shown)


gwbres:
@MrCry0 thank you for diving in my rewrite of the Hatanaka algorithm, although it is simple on paper, it is quite complex in the implementation details. I think next release will take it to yet another level, mostly thanks to you 👍 

another dimension of improvements would be to double check unecessary memory allocation and replace them with static/pre allocation (I tried to keep this to the minimum, it should be close to OK). But the lib also has a several constants that are not coded as such. Mostly when "blanking" or dealing with CRINEX compression/decompression trailing, spacing and blankings. This is pure performance improvements and will not change the behavior. It has been like this for 5+ years now :) the current performances are already quite good. If you are not familiar with this, you can find the results of the performance benchmarks, actually in the CI/CD runs, [right here](https://github.com/nav-solutions/rinex/actions/runs/34199913456). In otherwords, for every "test run" of the library, we get a glimpse of how it performs and we can compare releases to releases.

gwbres (approved):
Thank you for spotting this issue, those are very dug deep and hard to find 👍 
I'm looking forward to next release already 🚀

MrCry0:
@gwbres No problem, and thank you for this useful project. I was just trying to get my project (https://github.com/MrCry0/rinexfetch) working and realized I needed to fix this one first :)

PR #442, opened 2026-09-07
navigation: write records that parse back in every revision

MrCry0:
## Summary

Every navigation file written by the library was corrupt or lossy in some way. Four commits:

- `navigation: write ephemeris orbit lines the parser reads back`: orbit lines were indented by three blanks for every revision, while the RINEX 3 and 4 parsers skip four, so every written RINEX 3/4 file parsed back to shifted values. SBAS transmission times (kept as the `week` orbit item) were written as zero, and fields absent from the record were written as zero instead of blank. The navigation comparison of the test toolkit only checked key presence, which is why the round trips passed; it now compares the frames.
- `header: write the LICENSE OF USE and DOI lines`: parsed but never written (BRD400DLR carries a DOI).
- `navigation: write the RINEX 4 STO, EOP and ION records`: only ephemerides were written; other records got a header line without termination and no body, so records ran together and the file could not be parsed back. Records sharing epoch, vehicle and frame type (Galileo INAV/FNAV) were collapsed to one. The STO UTC identifier, discarded by the parser although the field exists, is kept.
- `navigation: write RINEX 2 records with I2 fields and a D exponent`: `01 23 01 01 00 00` instead of ` 1 23  1  1  0  0`, year as `yyyy - 2000`, and the `E` exponent of RINEX 3 where RINEX 2 files use `D`.

## Test plan

- Strict formatting round trips (parse, write, reparse, frames equal) of every navigation file of the data folder: `production::nav_v2`, `nav_v3`, and `nav_v4` which was ignored and is enabled.
- Every STO, EOP and ION record of the two RINEX 4 files is written back byte for byte (case of the exponent letter and trailing blanks aside; the STO transmission time is not stored and not compared).
- Unit tests for the ephemeris layout (RINEX 2 and 3, SBAS), the D exponent, the RINEX 2 epoch line, the RINEX 4 record header and the header lines.
- `cargo test --all-features` green at every commit (210 to 214 tests).

## Notes

- Not covered: writing a record in a revision other than the parsed one (RINEX 4 records as RINEX 3 and the reverse). That needs message type assignment and belongs with the RINEX 4.02 navigation changes.
- The RINEX 4 parser panics on truncated lines (`split_at` on short input); the writer fix removes the trigger, the parser is not hardened here.
- The STO transmission time is not stored by `TimeOffset`; the reference time (seconds of week) is written in its slot.

### DIFF
--- a/src/header/formatting.rs
+++ b/src/header/formatting.rs
@@ -33,6 +33,15 @@ impl Header {
         }
 
         self.format_comments(w)?;
+
+        if let Some(license) = &self.license {
+            writeln!(w, "{}", fmt_rinex(license, "LICENSE OF USE"))?;
+        }
+
+        if let Some(doi) = &self.doi {
+            writeln!(w, "{}", fmt_rinex(doi, "DOI"))?;
+        }
+
         self.format_rinex_dependent(w)?;
 
         if let Some(rcvr) = &self.rcvr {
--- a/src/navigation/ephemeris/formatting.rs
+++ b/src/navigation/ephemeris/formatting.rs
@@ -33,31 +33,39 @@ impl Ephemeris {
             },
         };
 
-        // starts with (clock_bias, drift, rate)
+        // starts with (clock_bias, drift, rate).
+        // SBAS messages carry the transmission time (seconds of week)
+        // in the drift rate slot, stored as the "week" orbit item.
         // epoch has already been buffered
+        let third = if sv_constellation == Constellation::SBAS {
+            self.get_orbit_f64("week").unwrap_or(self.clock_drift_rate)
+        } else {
+            self.clock_drift_rate
+        };
+
         write!(
             w,
             "{}{}{}",
             NavFormatter::new(self.clock_bias),
             NavFormatter::new(self.clock_drift),
-            NavFormatter::new(self.clock_drift_rate),
+            NavFormatter::new(third),
         )?;
+
+        // orbit lines: three leading blanks in RINEX 2, four from RINEX 3 on.
+        // Fields absent from the record are left blank, as the parser
+        // read them.
+        let padding = if version.major < 3 { "   " } else { "    " };
+        const BLANK: &str = "                   ";
 
         // following standard specs
         let data_fields = &standard_specs.items;
-        for i in 0..data_fields.len() {
-            if let Some(value) = self.get_orbit_f64(data_fields[i].0) {
-                if i % 4 == 0 {
-                    write!(w, "\n   {}", NavFormatter::new(value))?;
-                } else {
-                    write!(w, "{}", NavFormatter::new(value))?;
-                }
-            } else {
-                if i % 4 == 0 {
-                    write!(w, "\n   {}", NavFormatter::new(0.0))?;
-                } else {
-                    write!(w, "{}", NavFormatter::new(0.0))?;
-                }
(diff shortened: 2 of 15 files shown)


gwbres:
Wow that was very far from working, I did not realize that V4 formatting was that far

src/header/formatting.rs
```diff
@@ -33,6 +33,15 @@ impl Header {
...
         self.format_comments(w)?;
+
+        if let Some(license) = &self.license {
+            writeln!(w, "{}", fmt_rinex(license, "LICENSE OF USE"))?;
+        }
+
+        if let Some(doi) = &self.doi {
+            writeln!(w, "{}", fmt_rinex(doi, "DOI"))?;
+        }
+
```
gwbres, line 44: 👍

src/navigation/ephemeris/formatting.rs
```diff
@@ -33,31 +33,46 @@ impl Ephemeris {
...
-        // starts with (clock_bias, drift, rate)
+        // starts with (clock_bias, drift, rate).
+        // SBAS messages carry the transmission time (seconds of week)
+        // in the drift rate slot, stored as the "week" orbit item.
         // epoch has already been buffered
+        let third = if sv_constellation == Constellation::SBAS {
+            self.get_orbit_f64("week").unwrap_or(self.clock_drift_rate)
+        } else {
+            self.clock_drift_rate
+        };
```
gwbres, line 44: 👍 good catch. I did that properly on the parsing side, this will be useful

src/navigation/ephemeris/formatting.rs
```diff
@@ -109,19 +126,71 @@ mod test {
...
+
+        assert!(utf8.starts_with(
+            "-1.000000000000E-04-2.000000000000E-11 0.000000000000E+00
+     1.000000000000E+00 2.000000000000E+00 3.000000000000E+00                   
+     5.000000000000E+00"
+        ));
+    }
+
+    #[test]
+    fn sbas_ephemeris_formatting() {
```
gwbres, line 163: Thank you for adding more tests. Even the simplest tests are useful, they garantee non-regression

src/tests/formatting/header.rs
```diff
@@ -295,3 +295,24 @@ fn crinex_gal_v4_header_formatting() {
         ]),
     );
 }
+
+#[test]
+fn header_license_doi_formatting() {
```
gwbres, line 300: 👍

src/tests/formatting/nav.rs
```diff
@@ -0,0 +1,121 @@
+//! Navigation RINEX formatting, compared to the original files.
+use crate::{navigation::NavFrameType, prelude::Rinex};
+
+use std::{collections::HashMap, fs::read_to_string, fs::remove_file};
+
+/// Splits a RINEX 4 navigation record section into blocks, one per
+/// "> XXX" record, keyed by the record header, the epoch and, for STO
+/// records, the time system pair. Exponent letters are uppercased and
+/// trailing blanks trimmed on every line.
+fn v4_record_blocks(content: &str) -> HashMap<String, Vec<String>> {
```
gwbres, line 10: Thank you for going that far and improving our test ecosystem, this allows to write compelling and advanced tests, as you have already used elsewhere. The Observation ecosystem is the most "advanced", this will be very useful and allow other format, mostly navigation, to reach the same quality level

gwbres (approved):

PR #444, opened 2026-09-07
hatanaka: fix CRINEX V1 decompression

MrCry0:
## Summary

CRINEX 1 (RINEX 2) files decompressed to misaligned records: the strict signal by signal comparison with their RINEX 2 counterparts (`testbench_v1`, disabled on `main`) failed on every pair. Two commits:

- `hatanaka: reset the kernels of vehicles absent from the previous epoch`: in CRINEX a vehicle missing from an epoch restarts from scratch when it comes back. Neither side did that: the compressor coded returning vehicles against kernels CRX2RNX has discarded, the decompressor kept the old flags of returning vehicles when they were published as blanks. Both sides now drop the kernels of the vehicles missing from an epoch. The CRINEX 3 files written from the test resources are accepted by CRX2RNX 4.1.0 (they already were; CRINEX 1 output is rejected by CRX2RNX with or without this change, the RINEX 2 epoch line layout of the compressor is a separate, pre-existing issue and `crinex1` stays ignored).
- `hatanaka: fix CRINEX V1 decompression`: observation lines were tokenized with ad hoc rules that mishandled runs of missing observations, their flags and single digit values; the V1 epoch description emitted one continuation line too many from 11 vehicles on and a termination the caller adds again (a blank line the RINEX 2 parser reads as five missing observations); stale vehicle identifiers of a longer earlier epoch trailed the description; the record parser fed compressed lines that produce no output to the RINEX 2 parser as observation lines (which ran past the SV list and panicked in the SV parser on an empty string).

## Test plan

- `testbench_v1` enabled: AJAC3550, KOSG0010, aopr0010, delf0010, npaz3550, wsra0010 and zegv0010 decompress to the same record as their RINEX 2 files, signal by signal (barq071q is left out, its RINEX model holds a single epoch).
- The raw AJAC3550 expectation is regenerated from the CRX2RNX 4.1.0 output of the file; the zegv0010 expectation loses the blank line of the old extra termination.
- `cargo test --all-features` green at both commits (209 and 210 tests).

## Notes

- Compressed values are still decompressed with `NumDiff::decompress` returning a plain integer; once #441 merges the call site takes a `?`.
- The delf0010 clock offset extract of nav-solutions/data#6 will exercise this path with receiver clock offsets (the clock offset PR).


src/tests/crinex/decompression/mod.rs
```diff
@@ -99,195 +100,29 @@ use std::{
...
-//         // TODO unlock this
-//         generic_observation_comparison(&dut, &model);
-
-//         // clean up & exit
-//         let _ = std::fs::remove_file(filename);
-//     }
-// }
+/// Every CRINEX 1 file of the data folder decompresses to the record of
+/// its RINEX 2 counterpart, signal by signal.
+#[test]
```
gwbres, line 105: If that proves to stand while preserving RINEX3/CRINEX3 , this is very good achievement ! 

Although it was achieved a few years ago, I can't remember why this test started to fail and had to be ignored. Hopefully we will be able to preserve it from now on

gwbres (approved):

gwbres:
Thank you for rebasing, very good work once again 👍

gwbres (approved):

PR #445, opened 2026-09-07
observation: fix the signal validity checks

MrCry0:
## Summary

Two validity checks of `SignalObservation` did not do what their documentation says:

- `is_ok()` rejected every signal without an SNR indication (the default SNR is the weakest level, never "strong"), although a missing SNR is documented as accepted. A missing SNR now passes; a reported one must still be strong.
- `is_ok_snr()` tested the LLI flags with `intersects()` against `OK_OR_UNKNOWN`, the empty flag set, so it returned false for every signal, whatever its SNR.

One commit, with a unit test covering signals without flags, strong and weak SNR, sane LLI flags and a lock loss.

## Test plan

- `cargo test --all-features` green (210 tests).


gwbres (approved):
LGTM

PR #443, opened 2026-09-07
navigation: RINEX 4.02 support

MrCry0:
## Summary

RINEX 4.02 navigation support, rebased on `main` after #442 merged. #446 (revision conversion) and #447 (#426 fix) are stacked on this branch and must be merged after it.

On `main`, the specification examples file (34 records, 30 keys) parsed to 15 keys. Record by record:

- `navigation: keep the STO time system pair as written`: the GLONASS and NavIC pairs (GLUT, GLGP, IRUT, IRGP, BDGL) were rejected, dropping the STO records and the TIME SYSTEM CORR header lines carrying them. hifitime has no GLONASS nor NavIC time scale: they map to UTC and GPST, and the four letter code is kept in a new `TimeOffset::time_system` field so it writes back as read.
- `navigation: parse the RINEX 4.02 record header line`: message types L1NV, L1OC, L3OC, LXOC; optional subtype (WIDE, JAPN, KLOB, NEQN) kept in a new `NavKey::subtype` field; blank PRN headers (`> STO E   IFNV`) as PRN 0; NavIC records in GPST (no timescale in gnss-rs, so every NavIC record, RINEX 3 included, was rejected before).
- `navigation: keep orbit fields aligned after a blank field`: a blank field followed by omitted trailing fields on the same line shifted every following field (GPS LNAV line 5 with the blank L2 codes).
- `navigation: add NavIC LNAV and L1NV ephemeris definitions` (Tables A30, A31; there was no NavIC entry at all), `add GLONASS L1OC and L3OC ephemeris definitions` (Tables A16, A17), `add the optional CNAV and CNV2 flags field` (Tables A10, A11, A20, A21).
- `navigation: add the NavIC L1NV and GLONASS CDMA ionosphere models` (Tables A38 to A40), parsed and written back.
- `observation: stop the RINEX 2 SV list at a blank slot`: blank SV slots before a RINEX 2 clock offset panicked the SV parser; carries the data submodule bump to 209bfbd, which brings the delf0010 clock offset pair exercising it (nav-solutions/data#6) and the RINEX 4.02 examples file of the next commit (nav-solutions/data#7).
- `tests: cover the RINEX 4.02 navigation specification examples`.
- `carrier: add the GPS R attribute and NavIC L1 observation codes` (RINEX 4.01).

## Test plan

- Unit tests on the specification example records for every new message, model and pair, plus the alignment case.
[...]

PR #446, opened 2026-09-08
navigation: write records in a revision other than the parsed one

MrCry0:
> **WARNING: this PR MUST be merged AFTER #443.** It is stacked on `m3/rinex-4.02-navigation`: the ten commits of #443 show in the diff until it merges, only the last four commits belong here.

## Summary

Writing a navigation record in a revision other than the parsed one failed or mislabeled the messages. Reproduced on `main`:

- both RINEX 4 files of the data folder written as 3.05 fail with `MissingNavigationStandards`: the first CNAV record aborts the whole file, taking every legacy ephemeris with it. The rapid multi-GNSS products (BRD400DLR) mix GPS LNAV with CNAV and CNV2, so a GPS-only downconversion to RINEX 3 was impossible;
- a RINEX 3 file written as 4.00 files every ephemeris as `LNAV` (`> EPH R19 LNAV`, `> EPH C05 LNAV`).

Four commits (the develop counterparts of the last three were part of #438):

- `navigation: tell the legacy messages from the RINEX 4 only ones`: `NavMessageType::is_legacy(constellation)`, the predicate the writer and the tests share: the legacy message of each constellation (LNAV, FDMA for GLONASS, INAV/FNAV for Galileo, D1/D2 for BeiDou, SBAS for SBAS vehicles) has a RINEX 2/3 representation, the modern messages and the STO, EOP and ION types do not. Public, since a user filtering a record before a downconversion needs the same rule.
- `navigation: write records in a revision other than the parsed one`: written as RINEX 2 or 3, keep the ephemerides of the legacy messages laid out with the RINEX 3 definitions, skip the modern messages (CNAV, CNV1 to CNV3, L1NV, L1OC, L3OC), the STO, EOP and ION records, and the satellites of other constellations in a single constellation RINEX 2 file. Written as RINEX 4, give the ephemerides parsed from RINEX 2 or 3 their message type: FDMA for GLONASS, SBAS for SBAS, INAV or FNAV for Galileo from the data source (Table A13), D1 or D2 for BeiDou from the GEO PRN range (BDS ICD). The BeiDou group delays, named differently by the RINEX 3 and 4 definitions, are written under either name.
- `navigation: fail when no frame can be written in the target revision`: `FormattingError::NoRepresentableFrame` when the record is not empty and none of its frames is representable, instead of a file with a header and no record.
- `tests: cover the GPS only RINEX 4 to 3 navigation conversion`: the BRD400DLR product filtered to GPS.

## Test plan

- KMS300DNK and BRD400DLR written as 3.05 keep exactly their legacy ephemerides, each field equal to the RINEX 4 value (records sharing epoch and vehicle, such as Galileo INAV/FNAV pairs, collapse to one RINEX 3 key: known limitation of the RINEX 3 layout).
- BRD400DLR filtered to GPS written as 3.05: every LNAV ephemeris kept, CNAV and CNV2 dropped.
- AMEL (RINEX 3) written as 4.00 gets the expected message types with every field preserved, and written back as 3.05 gives the original record. CBW (mixed RINEX 3) written as a 2.11 GPS file keeps the GPS ephemerides only. A record reduced to CNAV/CNV2 messages written as 3.05 returns `NoRepresentableFrame`.
- Unit test of `is_legacy` over every message type.
- `cargo test --all-features` green at every commit (240 to 245 tests).
- Checked with the rinexfetch tool (MrCry0/rinexfetch, `fix/nav-writer-negative-field-indent`) built against this branch: its GPS-only RINEX 3 output of the BRD400DLR product now succeeds (428 ephemerides), where it hit `MissingNavigationStandards` before.

## Notes

- New `FormattingError::NoRepresentableFrame` variant and `NavMessageType::is_legacy` method.

### DIFF
--- a/src/navigation/message.rs
+++ b/src/navigation/message.rs
@@ -1,7 +1,7 @@
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-use crate::prelude::ParsingError;
+use crate::prelude::{Constellation, ParsingError};
 
 #[cfg(doc)]
 use crate::bibliography::Bibliography;
@@ -92,6 +92,24 @@ impl std::str::FromStr for NavMessageType {
     }
 }
 
+impl NavMessageType {
+    /// Returns true if this message, broadcast by `constellation`, has a
+    /// representation in RINEX 2 and 3: the legacy message of each
+    /// constellation, which those revisions do not name. The modern
+    /// messages (CNAV, CNV1 to CNV3, L1NV, L1OC, L3OC) only exist in
+    /// RINEX 4, as do the STO, EOP and ION message types.
+    pub fn is_legacy(self, constellation: Constellation) -> bool {
+        match self {
+            Self::LNAV => true,
+            Self::FDMA => constellation == Constellation::Glonass,
+            Self::INAV | Self::FNAV => constellation == Constellation::Galileo,
+            Self::D1 | Self::D2 => constellation == Constellation::BeiDou,
+            Self::SBAS => constellation.is_sbas(),
+            _ => false,
+        }
+    }
+}
+
 impl std::fmt::Display for NavMessageType {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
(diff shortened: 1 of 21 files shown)


src/navigation/message.rs
```diff
@@ -81,6 +93,89 @@ impl std::fmt::Display for NavMessageType {
...
     }
 }
+
+/// Navigation message subtype, introduced by RINEX 4.02 as an optional
+/// fourth field of the navigation record header line (section 5.4.1).
+/// Subtypes are only defined for a few ION messages (Table 25) and
+/// distinguish otherwise identical records.
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum NavMessageSubtype {
```
gwbres: Is this not a duplicate (and possible update) of [the existing KbRegionCode](https://docs.rs/rinex/latest/rinex/navigation/enum.KbRegionCode.html) ? 

As I'm re-reading the specs.. KbRegionCode exists only in V4 Navigation, for ION message frames, for BeiDou CNVX or GPS satellites.

Does that mean that V4.02 introduces yet another flag (possibly _before_) all of that and the previous still applies ?

